PR #11, opened 2015-01-30
Update libraries and JSX for current versions 1/23/2015

justin808:
* Required a few JSX changes.
* Don't use import for commonjs modules
* key changed to eventKey for NavItem
* Use different names for the the native input HTML elements compared to
  the react-bootstrap ones
* Could not update the sass-loader without errors
* Moved all Sass related parts of webpack to the webpack.hot.config.js
  as Rails version does not use the Webpack sass handlers.
* Changed webpack.rails.config.js so that the loader for JSX does not
  use react-hot
* Added the woff2 pattern for the file loader.
* Changed from React.renderComponent to React.render in example.jsx.


justin808:
@mbreining @kimagure Any comments? I'll try to get the sass-loader updated next.

kimagure:
You can import CommonJS modules easily, just requires that you do `import * as YOURNAME from 'module'`

I admit it's not the best thing ever, but it'll work enough.

mbreining:
@justin808, the code in ```update-to-latest``` is deploying to Heroku fine but is not functional (none of the forms are showing up). The same issue exist on ```master```. I'll have to look into it.

mbreining:
You can merge as far as I'm concerned.

justin808:
@mbreining I rebased master onto my branch and did a push -f.

Issue #10, opened 2015-01-28
Move package.json to inside webpack directory


justin808:
We can have a top level package.json that loads the /webpack/package.json.

This keeps the top level tidier.

From Heroku support:

I agree - as long as an extra root-level package.json is okay with you, it's probably a better bet to just use the officially supported buildpack. You can trigger sub-installs like this (in package.json):

```
  "scripts": {
    "postinstall": "cd ./client && npm install && cd ../server && npm install"
   }
```

Also, I'd love to get your feedback on how that works with the upcoming version of the Node buildpack (codenamed "yoga"):
https://github.com/heroku/heroku-buildpack-nodejs/tree/yoga
If you decide to try it out, please let me know what you think!
Thanks,
Hunter

bdefore:
I decided to try this out and it works without many changes. My Procfile.dev became:
```
webpack: cd ./client && webpack -w --config ./client/webpack.rails.config.js
```
(not that i renamed 'webpack' dir to 'client')

Overall I'm much happier with the separation, with dependencies defined and installed relative to client. It also has the benefit of making it more straightforward for porting this process over to an existing rails setup, since you could just copy over the 'client' directory, make a few Procfile customizations and you're good to go.

justin808:
@bdefore Awesome. One advantage of the name client is that if you use zsh and type webpack to switch directories, well, webpack is a exe. If you can submit a PR, I'll get that integrated. Otherwise, @mbreining  or I will try to change it shortly.

justin808:
I'm creating this change. @bdefore I'll notify you so that you might be able to comment on the PR.

justin808:
@bdefore If you get a chance, please review Pull request #19.

bdefore:
@justin808 i've left a couple notes but overall looks like what i'd expect. you'll also want to update your blog post references to 'webpack' directory since that may be how people come to find your tutorial.

justin808:
Closed by #19.

PR #19, opened 2015-03-23
Move package json inside webpack directory

justin808:
1. Changed everything client side to be under /client directory
2. Moved node_modules to /client directory
3. Updated libraries including switching to babel.

Addresses:
Move package.json to inside webpack directory #10


mbreining:
Looks good.

justin808:
This code is running on heroku right now.

client/assets/javascripts/CommentBox.jsx
```diff
@@ -5,12 +5,12 @@ var React = require('react/addons');
...
-var Nav = require('react-bootstrap/Nav')
-var NavItem = require('react-bootstrap/NavItem')
+var Input = require('react-bootstrap/lib/Input');
+var Button = require('react-bootstrap/lib/Button');
+var Row = require('react-bootstrap/lib/Row');
+var Col = require('react-bootstrap/lib/Col');
+var Nav = require('react-bootstrap/lib/Nav')
+var NavItem = require('react-bootstrap/lib/NavItem')
 
 var marked = require("marked");
```
bdefore: Is this change intended? On another project, I have been using these paths without specifying lib.

client/webpack.rails.config.js
```diff
@@ -7,18 +7,19 @@
...
 config.entry.push("./scripts/rails_only"); // rails specific assets
-config.output = { filename: "webpack-bundle.js",
+config.output = { filename: "client-bundle.js",
                   path: "../app/assets/javascripts" };
 config.externals = { jquery: "var jQuery" }; // load jQuery from cdn or rails asset pipeline
 config.module.loaders.push(
-  { test: /\.jsx$/, loaders: ["es6", "jsx?harmony"] },
-  // Next 2 lines expose jQuery and $ to any JavaScript files loaded after webpack-bundle.js
+  { test: /\.jsx$/, exclude: /node_modules/, loader: 'babel-loader'},
+  // Next 2 lines expose jQuery and $ to any JavaScript files loaded after client-bundle.js
```
bdefore: I believe you might want some .jsx from node_modules to be included? For example, I use the react-paginate library.

client/assets/javascripts/CommentBox.jsx
justin808: Yes, part of updating react-bootstrap.

client/webpack.rails.config.js
justin808: That's a valid point. I think the key is to not have babel parsing every one. Maybe it makes sense to create a separate test for that?

justin808:
@bdefore Thanks for the feedback!

PR #27, opened 2015-04-05
Flux integration

mbreining:
@justin808 Please review. I added support for Flux (https://github.com/goatslacker/alt). The functionality remains the same.

Note that the latest master has been merged into this branch.

Also note that the alt library does not get installed by `npm install` although it is defined in package.json. I suspect there is a dependency conflict somewhere, although I would expect npm to error out. I didn't get a chance to investigate so for now I'm installing it manually `cd client && npm install alt`.


client/assets/javascripts/components/Comment.jsx
```diff
@@ -0,0 +1,27 @@
...
+  propTypes: {
+    author: React.PropTypes.string.isRequired,
+    text: React.PropTypes.string.isRequired
+  },
+
+  render() {
+    const rawMarkup = marked(this.props.text);
+    return (
+      <div className="comment">
+        <h2 className="commentAuthor foobar">
```
justin808: foobar?

client/server.js
```diff
@@ -30,7 +30,7 @@ server.app.post('/comments.json', function(req, res) {
   console.log('Just got done with nap!');
   comments.push(req.body.comment);
   res.setHeader('Content-Type', 'application/json');
-  res.send(JSON.stringify(req.body.comment));
+  res.send(JSON.stringify(comments));
```
justin808: send back all comments? OK I guess if we knew the list was limited.
I have it coded to send back just the comment that's saved.

client/webpack.common.config.js
```diff
@@ -5,7 +5,7 @@ var path = require('path');
 
 module.exports = {
   context: __dirname, // the project dir
-  entry: ['./assets/javascripts/example'],
+  entry: ["./assets/javascripts/App"],
```
justin808: better

fredericgrati:
I have the following error when I tried to go on localhost:4000
```
Uncaught Error: Cannot find module "../actions/FormActions"
```
Everything works fine on localhost:3000.

Thx for your alt integration. Awesome work !

mbreining:
@fredericgrati This PR is no longer valid. I will close it shortly. Check out PR #28 instead. Thanks.

PR #34, opened 2015-05-09
Enable rails logging with Foreman

lifehackett:
I was finding that more often than not my rails logs were being swallowed by Foreman.  This should fix that.

### DIFF
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,3 +1,4 @@
 web: rails s -p 4000
 client: cd client && $(npm bin)/webpack -w --config webpack.rails.config.js
 hot: cd client && node server.js
+log: tail -f log/development.log


justin808:
@lifehackett, this depends on your logging setup. On my projects, which I believe is the Rails default, development.rb typically logs to stdout. Can you paste an example here of a log message in this project that you prefer to see with your setup?

lifehackett:
I am just using the default setup as well.  If I run `rails s` in its own terminal window, I get all sorts of useful logging info in the terminal about http requests, db reads/writes, etc... all stuff that comes baked in to Rails.  Using foreman, I would only get these log messages ~10% of the time until I added the config in this PR

Example:
```
10:35:32 web.1    | Started GET "/api/v1/contacts" for 127.0.0.1 at 2015-05-11 10:31:40 -0400
10:35:32 web.1    | Processing by Api::V1::ContactsController#index as JSON
10:35:32 web.1    |   Contact Load (0.3ms)  SELECT "contacts".* FROM "contacts"
```

justin808:
Strange, it shouldn't be 10% of the time. Maybe it's due to the log level?

lifehackett:
Yes, I thought it was strange too.  I tried the log level but that didn't improve things.  If no one else is experiencing this problem then feel free to close this PR.  I found it very frustrating until I found this solution so thought it might benefit others.

justin808:
@lifehackett There's always a reason. The question is whether it's worth the time to figure it out. I've often been stubborn at not letting stuff like this slide by. I always want to know. My skills are where they are today from this attitude.

justin808:
I'm going to close this for now. Please re-open if you thein we should include this.

milushov:
@lifehackett 
>  If no one else is experiencing this problem
I have same problem :) exactly as you described

justin808:
@milushov 

By any chance do you have the heroku gem running? that will send all logs to stdout.

milushov:
@justin808 nope, this gem don't event installed

PR #44, opened 2015-07-14
Server rendering

justin808:


app/assets/javascripts/components.js
```diff
@@ -0,0 +1,2 @@
+//= require_tree ./components
```
justin808: There was no directory underneath this one.

justin808:
This is roughly how we're doing it: https://github.com/justin808/react_server_side_rendering_webpack/tree/redux-on-rails-server

PR #45, opened 2015-07-15
Server rendering example

justin808:
WIP right now.


justin808:
See #59. Server rendering will be included with that change.

justin808:
Not going to do it this way. See https://github.com/shakacode/react_rails_server_rendering/

PR #46, opened 2015-07-17
Add npm Karma plugins and rake task for JavaScript testing

Johnnycus:
There are no project tests, just sample test.


lib/tasks/tests.rake
```diff
@@ -0,0 +1,5 @@
+desc "Run Karma tests"
+task :test do
```
justin808: Let's rename this file `js_tests.rake` and the task as well to `js_tests`.

We can have a `test` task that does both rspec and karma.

justin808:
We can get some inspiration on the React components to test here: 

https://github.com/react-bootstrap/react-bootstrap/blob/master/test/ButtonSpec.js

@Johnnycus this is a good start!

The goal is to provide a clean simple example of how we'd add JS tests to a project like this.

This project is in active development: https://github.com/gaearon/redux and it might be the right example.

PR #48, opened 2015-07-19
Update Readme

nikhilbaradwaj:
I believe that npm install command should be executed inside the client folder.


justin808:
@nikhilbaradwaj I've had no troubles with running npm i from the top level.

The top level package.json calls the client/package.json.

This functionality is critical for the Heroku deploy.

nikhilbaradwaj:
Oh I see. For some reason, it didn't work for me but it might have to do with my setup. Thanks.

PR #55, opened 2015-08-01
Add rake task that delete all the comments older than one day

Johnnycus:
Add rake task for heroku scheduler.
1. Install scheduler addon: `heroku addons:create scheduler:standard`.
2. Deploy task to heroku.
3. Run `heroku run rake rm_comments` for test. **It will delete all comments from project page.**
4. Open scheduler page `heroku addons:open scheduler`.
5. Set it like this: scheduler page

### DIFF
--- a/lib/tasks/scheduler.rake
+++ b/lib/tasks/scheduler.rake
@@ -1,0 +1,5 @@
+desc "Remove all comments older than one day"
+task :rm_comments => :environment do
+  puts "Removing comments older than one day"
+  Comment.delete_all(:created_at => Time.now.midnight - 1.year..Time.now.midnight - 1.day)
+end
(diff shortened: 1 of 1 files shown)


lib/tasks/scheduler.rake
```diff
@@ -0,0 +1,5 @@
+desc "Remove all comments older than one day"
+task :rm_comments => :environment do
+  puts "Removing comments older than one day"
+  Comment.delete_all(:created_at => Time.now.midnight - 1.year..Time.now.midnight - 1.day)
```
justin808: We use the ruby 1.9 has syntax `created_at:`

justin808: Why does 1 year back matter?

justin808:
We need to adjust the UI with some stylish syntax that suggests all comments older than 24 hours get deleted each day, and that markdown is supported.

lib/tasks/scheduler.rake
Johnnycus: 1 year is needed to delete all the previous comments, not only for the last day

justin808:
See #62. Already done.

PR #75, opened 2015-09-04
more justin808 -> shakacode conversions

bronson:
Author in both package.jsons is still justin808, wasn't sure there.  But the rest seem important.


justin808:
@bronson Sweet. Thanks so much! LOVE getting help from the community!

PR #76, opened 2015-09-05
Make localhost links in README clickable

bronson:
It's surprising to click a URL and end up at a different URL.


justin808:
Excellent!

PR #73, opened 2015-08-29
Use fetch instead of $.ajax

mapreal19:
- This closes #69

PR #79, opened 2015-09-05
Changed protect_from_forgery to use "with :null_session"

mapreal19:
This is the preferred value for APIs


mapreal19:
Closed as per discussion on #73

PR #82, opened 2015-09-07
Using react_on_rails gem for rendering components on the client

mapreal19:
- Paired with @samnang

### DIFF
--- a/app/views/pages/index.html.erb
+++ b/app/views/pages/index.html.erb
@@ -16,7 +16,6 @@
   </li>
 </ul>
 <hr/>
-<div id="content"></div>
-
-
-
+<div id="content">
+  <%= react_component('App') %>
+</div>
--- a/Gemfile
+++ b/Gemfile
@@ -37,6 +37,8 @@ gem "rails-html-sanitizer"
 # Use Unicorn as the app server
 gem "unicorn"
 
+gem "react_on_rails", "~> 0.1.1"
+
 gem "autoprefixer-rails"
 
 # Use Capistrano for deployment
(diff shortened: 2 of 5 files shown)


justin808:
@samnang  @mapreal19 , please let me know if this ready for merging and deployment.

app/views/pages/index.html.erb
```diff
@@ -16,7 +16,6 @@
   </li>
 </ul>
 <hr/>
-<div id="content"></div>
-
-
-
+<div id="content">
```
justin808: Do we need the wrapper for id "content" still?

app/views/pages/index.html.erb
```diff
@@ -16,7 +16,6 @@
...
+  <%= react_component('App') %>
```
justin808: OK -- we should either comment or explicitly say that server rendering is off.

I think the next PR should be two have 2 tabs that show with and without server rendering.

@alexfedoseev @samnang @skv-headless @mapreal19 @PlasticLizard thoughts?

alexfedoseev: I see no reasons why it can't be implemented (:

samnang: @justin808 I totally agree we will 2 separate tabs for client side rendering and server side rendering. The idea for this PR is for client side rendering.

app/views/pages/index.html.erb
samnang: Actually, we don't need it. But we haven't looked at if we need it for any styling.

samnang:
@justin808 everything is ready and work as what we have before. We've just switched to use the gem instead.

justin808:
Sweet. I'm going to look at this tonight, and probably deploy it!

justin808:
Please continue conversation on #84.

PR #84, opened 2015-09-08
react on rails gem setup

justin808:
@samnang @mapreal19 I just added a commit to make the hot reload server work again.

Did either of you try with http://localhost:3000?

@alexfedoseev How's this look?


mapreal19:
Looks good now. There is just some lint errors on `webpack.common.config.js`

samnang:
:+1:

PR #88, opened 2015-09-11
Move jQuery and jQuery-ujs under /client and npm

### LINKED ISSUE #51: Break out jquery from jquery-ujs: How to use webpack complete managed javascript library
hello @justin808 , thanks for your work.
i dont want to use javascript library gem. e.g. 
`jquery-rails`, `bootstrap-sass` in Gemfile.
follow you guide, i can import jquery from webpack's bundle file.
but when i remove `bootstrap-sass`, i can not `@import 'bootstrap'` in 
`app/assets/stylesheets`, could you give me a example about this?

dylangrafmyre:
Comment out jquery-rails gem
Remove require for jQuery and jQuery-ujs from application.js
Change required order of application.js client-bundle and es5-shim
Add jquery-ujs to package.json and npm shrinkwrap
Remove comments about jQuery from webpack.common.config.js
Remove jQuery loader expose from webpack.rails.config.js
Comment out jQuery import or external jQuery since it is served from
/client.
Move jQuery and jQuery-ujs expose to via require to rails_only.jsx entry point.
Add info to README about using jQuery with Rails and Webpack
This fixes #51

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,8 +18,7 @@ gem "coffee-rails"
 # See https://github.com/sstephenson/execjs#readme for more supported runtimes
 # gem "therubyracer", platforms: :ruby
 
-# Use jquery as the JavaScript library
-# gem "jquery-rails"
+# jquery as the JavaScript library has been moved under /client and managed by npm
 
 # Turbolinks makes following links in your web application faster. Read more: https://github.com/rails/turbolinks
 gem "turbolinks"
--- a/Gemfile
+++ b/Gemfile
@@ -19,7 +19,8 @@ gem "coffee-rails"
 # gem "therubyracer", platforms: :ruby
 
 # Use jquery as the JavaScript library
-gem "jquery-rails"
+# gem "jquery-rails"
+
 # Turbolinks makes following links in your web application faster. Read more: https://github.com/rails/turbolinks
 gem "turbolinks"
 # Build JSON APIs with ease. Read more: https://github.com/rails/jbuilder
--- a/Gemfile
+++ b/Gemfile
@@ -18,7 +18,9 @@ gem "coffee-rails"
 # See https://github.com/sstephenson/execjs#readme for more supported runtimes
 # gem "therubyracer", platforms: :ruby
 
-# jquery as the JavaScript library has been moved under /client and managed by npm
+# jquery as the JavaScript library has been moved under /client and managed by npm.
+# It is critical to not include any of the jquery gems when following this pattern or
+# else you might have multiple jQuery versions.
 
 # Turbolinks makes following links in your web application faster. Read more: https://github.com/rails/turbolinks
 gem "turbolinks"
--- a/README.md
+++ b/README.md
@@ -173,6 +173,14 @@ Be sure to see [assets.rake](https://github.com/shakacode/react-webpack-rails-tu
 The `webpack.rails.config.js` file generates client-bundle.js which is then included
 by the Rails asset pipeline.
 
+##jQuery with Rails and Webpack
+jQuery and jQuery-ujs are not required within `app/assets/javascript/application.js`
+and have been moved under`/client` and managed by npm. The modules are exposed as global via entry point
+`script/rails_only.jsx` by `webpack.rails.config.js`.
+
+Please refer to [Considerations for jQuery with Rails and Webpack](http://forum.railsonmaui.com/t/considerations-for-jquery-with-rails-and-webpack/344) for further info.
+
+
 ## Sass and images
 1. The Webpack server loads the images from the **symlink** of the
    `app/assets/images` directory.
--- a/README.md
+++ b/README.md
@@ -175,8 +175,8 @@ by the Rails asset pipeline.
 
 ##jQuery with Rails and Webpack
 jQuery and jQuery-ujs are not required within `app/assets/javascript/application.js`
-and have been moved under`/client` and managed by npm. The modules are exposed as global via entry point
-`script/rails_only.jsx` by `webpack.rails.config.js`.
+and have been moved under`/client` and managed by npm. The modules are exposed via entry point
+by `webpack.common.config.js`.
 
 Please refer to [Considerations for jQuery with Rails and Webpack](http://forum.railsonmaui.com/t/considerations-for-jquery-with-rails-and-webpack/344) for further info.
 
--- a/README.md
+++ b/README.md
@@ -178,6 +178,9 @@ jQuery and jQuery-ujs are not required within `app/assets/javascript/application
 and have been moved under`/client` and managed by npm. The modules are exposed via entry point
 by `webpack.common.config.js`.
 
+In `application.js`, it's critical that any libraries that depend on jQuery come after the inclusion 
+of the Webpack bundle, such as the twitter bootstrap javascript.
+
 Please refer to [Considerations for jQuery with Rails and Webpack](http://forum.railsonmaui.com/t/considerations-for-jquery-with-rails-and-webpack/344) for further info.
 
 
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -9,14 +9,11 @@
 //
 // Read Sprockets README (https://github.com/sstephenson/sprockets#sprockets-directives) for details
 // about supported directives.
-//
-//= require jquery
-//= require jquery_ujs
 
-//= require bootstrap-sprockets
-
-// Important to import jquery_ujs before rails-bundle as that patches jquery xhr to use the authenticity token!
-
+// Need to be on top to allow Poltergeist test to work with React.
 //= require es5-shim/es5-shim
+
+// It is important that generated/client-bundle must be before bootstrap since it is exposing jQuery and jQuery-ujs
 //= require generated/client-bundle
+//= require bootstrap-sprockets
 //= require turbolinks
--- a/client/npm-shrinkwrap.json
+++ b/client/npm-shrinkwrap.json
@@ -3,9 +3,9 @@
   "version": "1.1.0",
   "dependencies": {
     "babel-core": {
-      "version": "5.8.23",
+      "version": "5.8.24",
       "from": "babel-core@>=5.8.23 <6.0.0",
-      "resolved": "https://registry.npmjs.org/babel-core/-/babel-core-5.8.23.tgz",
+      "resolved": "https://registry.npmjs.org/babel-core/-/babel-core-5.8.24.tgz",
       "dependencies": {
         "babel-plugin-constant-folding": {
           "version": "1.0.1",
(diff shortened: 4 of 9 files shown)


Gemfile
```diff
@@ -19,7 +19,8 @@ gem "coffee-rails"
 # gem "therubyracer", platforms: :ruby
 
 # Use jquery as the JavaScript library
-gem "jquery-rails"
+# gem "jquery-rails"
```
justin808: Please delete, with some some comments as to why it's not there in the GemFile.

app/assets/javascripts/application.js
```diff
@@ -9,14 +9,11 @@
...
-//= require jquery_ujs
-
-//= require bootstrap-sprockets
-
-// Important to import jquery_ujs before rails-bundle as that patches jquery xhr to use the authenticity token!
 
+// Need to be on top to allow Poltergeist test to work with React.
 //= require es5-shim/es5-shim
+
+// It is important that generated/client-bundle must be before bootstrap since it is exposing jQuery and jQuery-ujs
```
justin808: :+1:

client/npm-shrinkwrap.json
```diff
@@ -825,272 +830,45 @@
       "from": "marked@>=0.3.5 <0.4.0",
       "resolved": "https://registry.npmjs.org/marked/-/marked-0.3.5.tgz"
     },
-    "node-libs-browser": {
```
justin808: How did this all change from this PR?

client/scripts/rails_only.jsx
```diff
@@ -8,3 +8,7 @@ require('es5-shim/es5-sham');
 // Due to issue: https://github.com/ariya/phantomjs/issues/12401
 // Phantomjs does not like promises
 require('es6-promise').polyfill();
+
+require('expose?jQuery!jquery');
+require('expose?$!jquery');
+require('expose?jquery-ujs!jquery-ujs');
```
justin808: I don't understand this line. `jquery-ujs` is not a valid js variable name, and it not something to expose like this.

What *is* important is that the code get invoked, which means that it needs to be in an entry point.

client/webpack.rails.config.js
```diff
@@ -12,7 +12,7 @@ config.output = {
 };
 
 // load jQuery from cdn or rails asset pipeline
-config.externals = {jquery: 'var jQuery'};
+// config.externals = {jquery: 'var jQuery'};
```
justin808: Let's remove all this.

client/scripts/rails_only.jsx
```diff
@@ -8,3 +8,7 @@ require('es5-shim/es5-sham');
 // Due to issue: https://github.com/ariya/phantomjs/issues/12401
 // Phantomjs does not like promises
 require('es6-promise').polyfill();
+
+require('expose?jQuery!jquery');
```
justin808: This should stay in the client/webpack.rails.config.js. No reason to put it here, is there?

client/webpack.rails.config.js
```diff
@@ -20,12 +20,7 @@ config.entry.push('./scripts/rails_only');
 // See webpack.common.config for adding modules common to both the webpack dev server and rails
 
 config.module.loaders.push(
-  {test: /\.jsx?$/, exclude: /node_modules/, loader: 'babel-loader'},
-
-  // Next 2 lines expose jQuery and $ to any JavaScript files loaded after client-bundle.js
-  // in the Rails Asset Pipeline. Thus, load this one prior.
-  {test: require.resolve('jquery'), loader: 'expose?jQuery'},
```
justin808: This should stay!

justin808:
A few need changes. Also, it's critical that you test this with `rails s -e production` so that you see if the CSRF stuff is working. The CSRF parts are not running during development, AFIK.

client/scripts/rails_only.jsx
dylangrafmyre: It did not work from within webpack.rails_only.config

client/webpack.rails.config.js
dylangrafmyre: Did not work

dylangrafmyre:
I was getting CSRF errors in development with jQuery being exposed in and not working from webpack.rails_only.config.js

client/npm-shrinkwrap.json
dylangrafmyre: Had to remove node_modules, npm install and  npm shrinkwrap to install jQuery-ujs npm package and to get npm shrinkwrap to finish without errors.  

So I assuming it updated itself from that process referencing the package.json and installed the latest version?

client/scripts/rails_only.jsx
dylangrafmyre: Ok, this is a entry point. How do you invoked it? I was just following the example from expose-loader, and thought it applied to the expose npm modules like jquery-ujs? 

It would be helpful if you provided a example of proper implementation?

alexfedoseev: @dylangrafmyre Looks like inclusion in an `entry` will invoke it:

```js
(function($, undefined) {
  // ...
})(jQuery)
```

Note: It should be included after `jQuery` is available.

dylangrafmyre: @alexfedoseev Thanks for the example! 

If I comment out ```require('expose?jquery-ujs!jquery-ujs')``` that I get: ```ActionController::InvalidAuthenticityToken (ActionController::InvalidAuthenticityToken)``` webserver error.

With it left as is, there are not CSRF issues and everything works properly.

So @justin808 @alexfedoseev why is inclusion in an ```entry``` preferred over the way is it exposed now? Thanks, for the education!

alexfedoseev: I guess this happens because `jquery-ujs` needs to be exposed to global scope to patch the global `$`. I'll take a closer look in an hour.

alexfedoseev: This way it works:

```js
/* client/webpack.common.config.js */

module.exports = {

  // the project dir
  context: __dirname,
  entry: ['jquery', 'jquery-ujs', './assets/javascripts/App'],

  resolve: {
    root: [
      path.join(__dirname, 'scripts'),
      path.join(__dirname, 'assets/javascripts'),
    ],
    extensions: ['', '.webpack.js', '.web.js', '.js', '.jsx', '.scss', '.css', 'config.js'],
  },
  module: {
    loaders: [

      // React is necessary for the client rendering:
      {test: require.resolve('react'), loader: 'expose?React'},
      {test: require.resolve('jquery'), loader: 'expose?jQuery'},
    ],
  },
};


/* client/scripts/rails_only.jsx */

require('es5-shim/es5-shim');
require('es5-shim/es5-sham');

// Due to issue: https://github.com/ariya/phantomjs/issues/12401
// Phantomjs does not like promises
require('es6-promise').polyfill();

// require('expose?jQuery!jquery');
// require('expose?$!jquery');
// require('expose?jquery-ujs!jquery-ujs');
```

dylangrafmyre: @alexfedoseev Great! That look so much better!

dylangrafmyre:
@justin808 Final Review please after making changes from @alexfedoseev comments about where to expose and add entry for jQuery within webpack.common.config.js.

Thank you @alexfedoseev!!!

Test pass and no CSRF issues in dev or production.

alexfedoseev:
@dylangrafmyre :+1:

Gemfile
```diff
@@ -18,8 +18,8 @@ gem "coffee-rails"
 # See https://github.com/sstephenson/execjs#readme for more supported runtimes
 # gem "therubyracer", platforms: :ruby
 
-# Use jquery as the JavaScript library
-gem "jquery-rails"
+# jquery as the JavaScript library has been moved under /client and managed by npm
+
```
justin808:
```
# It is critical to not include any of the jquery gems when following this pattern or else you might have multiple jQuery versions.
```

README.md
```diff
@@ -173,6 +173,14 @@ Be sure to see [assets.rake](https://github.com/shakacode/react-webpack-rails-tu
 The `webpack.rails.config.js` file generates client-bundle.js which is then included
 by the Rails asset pipeline.
 
+##jQuery with Rails and Webpack
+jQuery and jQuery-ujs are not required within `app/assets/javascript/application.js`
+and have been moved under`/client` and managed by npm. The modules are exposed via entry point
+by `webpack.common.config.js`.
+
```
justin808:
```
In `application.js`, it's critical that any libraries that depend on jQuery come after the inclusion of the Webpack bundle, such as the twitter bootstrap javascript.
```

Also, possibly the twitter bootstrap javascript should come from npm?

client/npm-shrinkwrap.json
justin808: Sounds fine.

client/webpack.common.config.js
```diff
@@ -6,12 +6,8 @@ module.exports = {
 
   // the project dir
   context: __dirname,
-  entry: ['./assets/javascripts/App'],
+  entry: ['jquery', 'jquery-ujs', './assets/javascripts/App'],
```
justin808: This makes sense. It's sort of ugly, but it makes sense.

justin808:
Couple small suggests. Please let me know if we should push to heroku.

dylangrafmyre:
@justin808 you want to push the branch to Heroku to test first before the merge?

dylangrafmyre:
Branch is live on Heroku. Looks good to me.

justin808:
Super. Let's merge and then push master.

client/webpack.common.config.js
alexfedoseev: It can be less ugly this way:

```js
entry: {
  app: './assets/javascripts/App',
  vendor: ['jquery', 'jquery-ujs']
}
```

But since all chunks will be squashed by assets pipeline, this will be useful only for the code appearance.

justin808: that's a neat option

I didn't know about that

let's go with that if it works!

alexfedoseev: I'll create PR today / tomorrow.

szyablitsky: [Documentation](http://webpack.github.io/docs/configuration.html#entry) says that Object instead of Array or String is used to get multiple outputs. I'm not sure this will work for one output.

justin808: I concur with @szyablitsky, according to his doc refs.

alexfedoseev: That's correct. But it can be easily linked in application.js and squashed by assets pipeline. Moreover if there will be multiple bundles, this needs to be done to avoid duplication of vendors libs in each bundle, so it will be example how to do it with webpack.

justin808: So two uses cases:

1. Simple one where we just have one bundle.
2. Slightly more complicated one creating a separate jquery, jquery-ujs

I think the performance of `webpack -w` will be a tiny bit better if we don't put the jquery in there every time.

Plus some rails pages might not need react stuff.

So I'd say we put this in as a good example of handling `jquery-bundle.js` as something in general for JS that needs to be available to both Rails and the other webpack bundles.

client/package.json
```diff
@@ -32,6 +32,7 @@
     "immutable": "^3.7.5",
     "imports-loader": "^0.6.4",
     "jquery": "^2.1.4",
+    "jquery-ujs": "^1.0.4",
```
justin808: @dylangrafmyre We need to update this to `^1.1.1-1`

client/webpack.common.config.js
alexfedoseev: Pushed: https://github.com/shakacode/react-webpack-rails-tutorial/pull/96

PR #91, opened 2015-09-14
Code coverage and badges

dylangrafmyre:
Add Coveralls.io gem "coveralls" to implement code coverage monitoring
Add status badges to README for Codeship ans Travis-CI builds, Coveralls code-coverage, and Gemansium gem dependency monitoring and security.

### DIFF
--- a/.coveralls.yml
+++ b/.coveralls.yml
@@ -1,0 +1,1 @@
+service_name: codeship
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ client-bundle.js.map
 .env
 node_modules
 npm-debug.log
+/coverage
 
 # Ignore bundle dependencies
 vendor/ruby
--- a/Gemfile
+++ b/Gemfile
@@ -12,7 +12,7 @@ gem "pg", group: :production
 gem "sass-rails"
 gem "bootstrap-sass", "~> 3.3.1"
 # Use Uglifier as compressor for JavaScript assets
-gem "uglifier"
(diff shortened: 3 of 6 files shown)


samnang:
Look good, but why do we need to setup both Codeship and Travis-CI? Should we pick just one?

dylangrafmyre:
@samnang a request was made to have travis-ci build in place. I added both badges to for both. Since we are both active on Codeship and Travis-CI.

Figured, you guys could pick one or leave both? It is free for open source. Codeship is the CI that is actively deploying to Heroku for the CD.

@justin808 your thoughts?

samnang:
I see. From my opinion, I think we could standardize our stack. Travis-CI could run multiple ruby versions, allows to cluster tests and run them in parallel, and it's pretty good for open source. CircleCI is also good for open source & private(free 1 container), and they also have good UI. Circle-CI, it has almost the same feature with Circle-CI and Travis-CI, but they have limit build(100 builds/month) for private project. They all support to deploy to Heroku.

I think we should only pickup one for a project because it makes the test status reliable and also quick to get to the finish status on PR's status check as well. And the integration with `Coveralls` should just be one.

I would choose Travis-CI for open source project because some open source projects like gems we want to tested in multiple ruby versions. And choose Circle-CI for private project because unlimited build.

justin808:
I agree with @samnang that we need only one active. However, I'd like this project to serve as an example, so I'm OK with both running so long as it doesn't slow us down a lot.

PR #94, opened 2015-09-16
Remove an unneeded section from the README

bronson:
No need to work hard installing capybara-webkit when it was removed by 07950bd08f63c

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,29 +250,6 @@ We have feature tests in /spec/features
 
 Run the tests with `rspec`.
 
-If you get errors when running rspec in that it can't find expected DOM elements, then you'll want to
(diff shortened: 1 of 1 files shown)


justin808:
@geoffevason @dylangrafmyre @samnang I'd agree with @bronson (thanks, btw!). I know some of our projects that we're using capybara-webkit. What's the recommendation on this one? I'd say we merge this one, and if we ever need to support both, we can add the instructions back later. Thanks again@bronson!

samnang:
@bronson @justin808 look good to me. I think we already have `poltergeist` in Gemfile, so it should be okay.

@bronson good catch :+1:

PR #100, opened 2015-09-22
Fix forum URL

StanBoyet:
There was a `www` at the top, which with the subdomain `forum` leads to a `ERR_NAME_NOT_RESOLVED`


justin808:
Thanks @StanBoyet !! :smile:

PR #106, opened 2015-09-29
Typo on driver type

mapreal19:


dylangrafmyre:
@mapreal19 Thanks for catching this!

PR #108, opened 2015-10-05
Unify use of jbuilder with server rendering

justin808:
* Change to use the same JSON for the server side load and the client
  side load of the comments.
* Reuse one jbuilder template for the list and single view of comments.
* Update the docs with how to do this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,33 +260,24 @@ cd client && npm run build:client && npm run build:server
 
 # JBuilder Notes
 There's a bunch of gotchas with using [Jbuilder](https://github.com/rails/jbuilder) to create the
-string version of the props to be sent to the react_on_rails_gem:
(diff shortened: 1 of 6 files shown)


justin808:
@samnang @alexfedoseev Please take a look.

README.md
```diff
@@ -257,6 +258,46 @@ npm install
...
+    generator_function: true, prerender: true) %>
+```
+
+However, if you try to set the value of the JSON string inside of the controller, then you will 
+run into several issues with rendering the Jbuilder template from the controller.
+See the notes in this the example code for app/controllers/pages_controller.rb.
+
+The two critical things:
+
+1. Use `render_to_string` to create string of JSON.
```
samnang: Do you want to give a note here.

PR #109, opened 2015-10-05
Add NODE_ENV variable to server build & Switch to react-transform-hmr

alexfedoseev:
* [x] Add `NODE_ENV` variable to server build
* [x] Switch to `react-transform-hmr`

/cc: @justin808

### DIFF
--- a/client/webpack.server.rails.config.js
+++ b/client/webpack.server.rails.config.js
@@ -1,5 +1,7 @@
 // Common webpack configuration for server bundle
 
+const webpack = require('webpack');
+
 module.exports = {
 
   // the project dir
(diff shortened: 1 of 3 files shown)


client/webpack.server.rails.config.js
```diff
@@ -8,13 +10,17 @@ module.exports = {
...
     filename: 'server-bundle.js',
     path: '../app/assets/javascripts/generated',
-
-    // CRITICAL for enabling Rails to find the globally exposed variables.
-    libaryTarget: 'this',
   },
   resolve: {
     extensions: ['', '.webpack.js', '.web.js', '.js', '.jsx', 'config.js'],
   },
+  plugins: [
```
justin808: Should we have a common setting for non-hot-reload? I think this node setting is duplicated in client and server. Maybe this is just fine as is.

justin808:
LGTM -- go ahead and merge if you think it's ready.

client/webpack.server.rails.config.js
alexfedoseev: `client.rails.config` inherits too much client stuff from `client.base.config`, so I'd keep it as is for readability sake until there will be more commons.

PR #115, opened 2015-10-08
Fix warnings of upgrading to React 0.14

josiasds:
- New required propTypes for `ReactCSSTransitionGroup`
- Provider component doesn't need to wrap a function anymore
- The `render` method now belongs to the `ReactDOM` lib

### DIFF
--- a/client/app/components/CommentForm.jsx
+++ b/client/app/components/CommentForm.jsx
@@ -225,8 +225,11 @@ const CommentForm = React.createClass({
     }
     return (
       <div>
-
-        <ReactCSSTransitionGroup transitionName="element">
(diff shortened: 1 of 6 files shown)


josiasds:
There's still one warning coming from `react_on_rails` gem:

screen shot 2015-10-08 at 17 15 26

justin808:
On the last deprecation warning, seems like we need different code in react_on_rails for the different versions of react.

justin808:
LGTM -- let's merge, and then let's fix the react_on_rails issue separately.

Since it's only a deprecation issue, then we can probably push to heroku.

PR #118, opened 2015-10-10
We need this as well for proper 0.14 server support.

rstudner:
Have to expose ReactDOMServer as well for the server rendering per 0.14 upgrade docs.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -207,7 +207,7 @@ GEM
     raindrops (0.15.0)
     rake (10.4.2)
     rdoc (4.2.0)
-    react_on_rails (0.1.8)
+    react_on_rails (1.0.0.pre)
       connection_pool
       execjs (~> 2.5)
       rails (>= 4.0)
@@ -353,7 +353,7 @@ DEPENDENCIES
   rails-html-sanitizer
   rails_12factor
   rainbow
-  react_on_rails (~> 0.1.8)
+  react_on_rails (~> 1.0.0.pre)
   rspec-rails
   rubocop
   ruby-lint
--- a/Gemfile
+++ b/Gemfile
@@ -39,7 +39,7 @@ gem "rails-html-sanitizer"
 # Use Unicorn as the app server
 gem "unicorn"
 
-gem "react_on_rails", "~> 0.1.8"
+gem "react_on_rails", "~> 1.0.0.pre"
 gem "therubyracer"
 
 gem "autoprefixer-rails"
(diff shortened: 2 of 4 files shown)


justin808:
We'll update this PR with the new gem version once we push that.

justin808:
@rstudner @dylangrafmyre Let's update the react_on_rail gem soon and push to heroku.

dylangrafmyre:
@justin808 what do you mean? The react_on_rails gem has already been updated to 1.0.0.pre as of last night for master.

justin808:
Cool! Just trying to keep up! Closing this one! Rocking!

rstudner:
Not sure why this was closed.  ReactDOM has been added to the client.base, but exposig ReactDOMServer still needs to go into the server.rails no?

(all this PR is for)

justin808:
@rstudner You are correct. I was under the impression that @dylangrafmyre did this.
We also need the update to the server one as well: https://github.com/shakacode/react_on_rails/blob/master/spec/dummy/client/webpack.server.js#L30

Please rebase these changes on top of master, and if the build passes, let's merge.

I'm curious as to how the builds worked if this was not there.

alexfedoseev:
> I'm curious as to how the builds worked if this was not there.

I think because old renderers are deprecated, not removed from core React package.

rstudner:
@alexfedoseev is correct.  ReactDOM & ReactDOMServer are not *required* for 0.14, you just get deprecation warnings.  Thus, the build will pass just fine without the change.

rstudner:
@justin808  rebased against master and pushed, build passed.

Gemfile.lock
```diff
@@ -353,7 +353,7 @@ DEPENDENCIES
   rails-html-sanitizer
   rails_12factor
   rainbow
-  react_on_rails (~> 0.1.8)
+  react_on_rails (~> 1.0.0.pre)
```
justin808: There's something wrong with this PR, as this is already on master: 

https://github.com/shakacode/react-webpack-rails-tutorial/blob/master/Gemfile.lock#L356

@rstudner Any ideas?

justin808:
@josiasds @dylangrafmyre This line is not right:
https://github.com/shakacode/react-webpack-rails-tutorial/blob/master/client/webpack.client.base.config.js#L32

We should not be using ReactDom from the server.

See examples in this directory: https://github.com/shakacode/react_on_rails/tree/master/spec/dummy/client

https://github.com/shakacode/react_on_rails#additional-steps-for-server-rendering-option-prerender-shown-below

justin808:
Sorry for the confusion @josiasds @dylangrafmyre, but my prior comment was incorrect. I just noticed that it is 'client.base' and not just 'base'. Do the line above is right.

PR #119, opened 2015-10-14
Reverse ports of Webpack and Rails server

### LINKED ISSUE #116: Consider reversing the 3000/4000 ports
Since Rails is a default of 3000, we should consider switching this:

https://github.com/shakacode/react-webpack-rails-tutorial/blob/master/client/server.js#L54

```javascript
server.listen(3000, 'localhost', function(err) {
  if (err) console.log(err);
  console.log('Listening at localhost:3000...');
});
```

https://github.com/shakacode/react-webpack-rails-tutorial/blob/master/Procfile.dev#L1

    web: rails s -p 4000

Anybody second me on this suggestion?

josiasds:
This fixes #116.

- Change Rails server port to default (3000).
- Change Webpack dev server port to 4000.
- Add simple server rendering tests

### DIFF
--- a/spec/requests/server_render_check_spec.rb
+++ b/spec/requests/server_render_check_spec.rb
@@ -1,0 +1,17 @@
+require "rails_helper"
+
+describe "Server Rendering" do
+  it "generates server rendered HTML if server rendering enabled" do
+    get root_path
+    html_nodes = Nokogiri::HTML(response.body)
+    expect(html_nodes.css("div#App-react-component-0").children.size).to eq(1)
+    expect(html_nodes.css("div#App-react-component-0 h2").text)
+      .to include("Comments")
+  end
+
+  it "generates no server rendered HTML if server rendering not enabled" do
+    get simple_path
+    html_nodes = Nokogiri::HTML(response.body)
+    expect(html_nodes.css("div#SimpleCommentScreen-react-component-0").children.size).to eq(0)
+  end
+end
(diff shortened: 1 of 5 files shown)


justin808:
@josiasds Can you throw in some tests for verifying that the server rendering worked:
https://github.com/shakacode/react_on_rails/blob/master/spec/dummy/spec/requests/server_render_check_spec.rb#L4

josiasds:
@justin808 Added some server rendering tests.

spec/requests/server_render_check_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+describe "Server Rendering" do
+  it "generates server rendered HTML if server rendering enabled" do
+    get root_path
+    html_nodes = Nokogiri::HTML(response.body)
+    expect(html_nodes.css("div#App-react-component-0").children.size).to eq(1)
+    expect(html_nodes.css("div#App-react-component-0 h2").text)
+      .to include("Comments")
+  end
+
+  it "generates no server rendered HTML if server rendering not enabled" do
```
justin808: @josiasds should we have fragment caching and server rendering enabled for the simple example? Let's have a link on the simple page to render with no server rendering. We can have the prerender be a query param for the get and if false, then no pre-rendering.

justin808:
@josiasds We can address my comments in your next PR.

josiasds:
@justin808 :+1:

PR #121, opened 2015-10-15
Set comment's author and text as not null and defaults as empty string

### LINKED ISSUE #110: Null vs "" for string fields
What’s the best practice for NULL vs “” for Rails apps with JS front ends (postgres)? Just ran into a thorny issue!

http://forum.shakacode.com/t/null-vs-for-rails-with-postgres-with-js-client-app/369 

@josiasds, @alexfedoseev, @samnang: Let's figure some best practice recommendation.

josiasds:
The last parameter of `change_column_null` only affects existing records, thus the new default value still needed to be set by `change_column_default`.

Fixes #110.

### DIFF
--- a/db/migrate/20151015160334_change_author_and_text_to_not_null.rb
+++ b/db/migrate/20151015160334_change_author_and_text_to_not_null.rb
@@ -1,0 +1,6 @@
+class ChangeAuthorAndTextToNotNull < ActiveRecord::Migration
+  def change
+    change_column_null(:comments, :author, false, "")
+    change_column_null(:comments, :text, false, "")
+  end
+end
--- a/db/migrate/20151015160035_set_default_for_author_and_text.rb
+++ b/db/migrate/20151015160035_set_default_for_author_and_text.rb
@@ -1,0 +1,11 @@
+class SetDefaultForAuthorAndText < ActiveRecord::Migration
+  def up
+    change_column_default(:comments, :author, "")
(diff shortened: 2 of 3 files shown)


josiasds:
@justin808 Ready for review.

db/migrate/20151015160334_change_author_and_text_to_not_null.rb
```diff
@@ -0,0 +1,6 @@
+class ChangeAuthorAndTextToNotNull < ActiveRecord::Migration
+  def change
+    change_column_null(:comments, :author, false, "")
+    change_column_null(:comments, :text, false, "")
+  end
+end
```
justin808: Why 2 migrations?

Does change_column_null handle migrating old records? We might need a couple lines of SQL to update the existing records. Then again, there probably aren't any, or else the UI would have crashed.

josiasds: Well, yes, that last parameter (empty string) of `change_column_null` replaces null values of *existing records*, so we don't need SQL statements here.

But it [still doesn't set the column's default](http://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-change_column_null) for new records, thus we do need a second migration to `change_column_default`.

justin808:
Thanks @josiasds

PR #123, opened 2015-10-15
meta tags manager

justin808:


client/app/utils/metaTagsManager.js
```diff
@@ -6,15 +6,10 @@ const MetaTagsManager = {
...
    */
   getCSRFToken() {
-    const metas = document.getElementsByTagName('meta');
-    for (let i = 0; i < metas.length; i++) {
-      const meta = metas[i];
-      if (meta.getAttribute('name') === 'csrf-token') {
-        return meta.getAttribute('content');
-      }
-    }
+    const token = Array.from(document.getElementsByTagName('meta'))
```
justin808: @josiasds Let's consider the lodash syntax from this.

Interesting find from @dylangrafmyre on the CI regarding `Array.from`.

josiasds: @justin808 I'm not sure we should add lodash only to replace `Array.from` by `_.toArray`.

justin808: You're right. We don't have lodash in this project.

```
_.find(document.getElementsByTagName('meta'), tag => tag.name === 'csrf-token')
```

s project.

I really like Lodash. I wouldn't suggest this otherwise. That being said, we don't *need* it give some features in ES6.

Because I want this example to be a real starter kit showing off Webpack/Npm/React integration, I think this is worth considering.

Let's see what @mapreal19, @alexfedoseev, @samnang, @dylangrafmyre Think about this one.

justin808:
@josiasds We can consider adding Lodash in a future PR.

josiasds:
@justin808 :+1:

client/app/utils/metaTagsManager.js
justin808: Even better from the lodash source, @jdalton:

```javascript
_.find(document.querySelectorAll('meta'), 'name', 'csrf-token')
```

justin808:
See https://github.com/shakacode/react-webpack-rails-tutorial/issues/124

alexfedoseev:
My decision making chart:

```
Can I use Immutable.js in this project?
            ⇙           ⇘
         Yes             No
       ⇙                    ⇘
Immutable.js               Lodash
```

justin808:
@alexfedoseev Why would you avoid lodash if you're using immutable.js?

alexfedoseev:
Because every Immutable instance already has all necessary methods in its prototype to handle the data transformations.

alexfedoseev:
Btw, in this case it's better to use not `npm i --save lodash` but `npm i --save lodash.find` to include only actually used lodash's method & reduce total build size.

jdalton:
`require('lodash/collection/find')`

alexfedoseev:
@jdalton :+1:

justin808:
@alexfedoseev do wepacks optimizations strip out unnecessary libs?

jdalton:
If you cherry-pick, then yes.

justin808:
@jdalton by cherry pick, you mean with the 

```
require('lodash/collection/find')
```

and can we include `npm i --save lodash` or do we need to do `npm i --save lodash.find`

Incidentally, this project is already including lodash as a dependency of a dependency.

alexfedoseev:
> and can we include `npm i --save lodash` or do we need to do `npm i --save lodash.find`

If we'll do `npm i --save lodash`, then we can:
```js
import _find from 'lodash/collection/find';
```

If we'll do `npm i --save lodash.find` then we can: 

```js
import _find from 'lodash.find';
```

Result will be the same. Only difference is if `lodash` will be installed, then we don't have to add stand-alone modules to `package.json` (like `lodash.find`). We can just cherry-pick them from `lodash/collection` dir.

@jdalton correct me if I'm wrong.

> Incidentally, this project is already including lodash as a dependency of a dependency.

I don't think it's a good idea to rely on such dependency, because we can't control it (it can be changed or removed from this package in the future versions). We should define this dependency implicitly. If lodash packages will be identical, then `webpack.optimize.DedupePlugin` should handle it.

rstudner:
Immutable.js > Lodash if we are talking about going into the future. (I love lodash, that isn't the point)

That being said, if we are going to use lodash, we 100% have to explicitly depend on it and not rely on a 3rd party library pulling it in.

jdalton:
> Immutable.js > Lodash if we are talking about going into the future.

ok

justin808:
:+1: :+1: :+1: 

@jdalton it would be nice if there was some way that we could use Lodash API's with immutable objects. We're prefixing the names of our immutable objects with `$$`, like `$$users` to indicate a immutable list of users. We ran into many bugs from forgetting to use the Immutable rather than the native JS APIs.

PR #125, opened 2015-10-21
Add current Git commit hash to view

robwise:


app/views/pages/index.html.erb
```diff
@@ -1,4 +1,9 @@
 <h2>Using React + Redux + Rails Backend (using the react_on_rails gem)</h2>
+<p>current Git commit:
+  <span id="current-git-commit-sha">
+    <%= @git_commit_sha %>
```
justin808: Let's make this clickable to go to the commit.

app/controllers/pages_controller.rb
```diff
@@ -1,6 +1,7 @@
 class PagesController < ApplicationController
   def index
     @comments = Comment.all
+    @git_commit_sha = GitCommitSha.current_sha
```
justin808: Let's make this a helper method so we don't have to set this to a variable.

app/views/pages/index.html.erb
robwise: @justin808 Good idea! I think I will also change it to use the shorter version.

app/models/git_commit_sha.rb
```diff
@@ -0,0 +1,23 @@
...
+    self.current_sha = ""
+  end
+
+  def self.retrieve_sha_from_git_folder
+    `git rev-parse HEAD 2>/dev/null`.to_s.strip
+  end
+
+  def self.retrieve_sha_from_env_var
+    env_var = ENV["DEPLOYMENT_SHA"]
+    env_var.blank? ? false : env_var
```
justin808: I prefer to use nil rather than having a value be a string or false.

app/models/git_commit_sha.rb
```diff
@@ -0,0 +1,23 @@
+class GitCommitSha
+  def self.current_sha
+    return @commit_sha unless @commit_sha.blank?
+    @commit_sha = (retrieve_sha_from_env_var || retrieve_sha_from_git_folder)
```
justin808:
```
def self.current_sha
  @commit_sha ||= retrieve_sha_from_env_var.presence || retrieve_sha_from_git_folder
end
```

spec/features/pages_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+  context "when env var is set" do
+    let(:sha) { "94d92356828a56db25fccff9d50f41c525eead5y" }
+    before do
+      ENV["DEPLOYMENT_SHA"] = sha
+      GitCommitSha.reset_current_sha
+    end
+    it_behaves_like "Git Commit SHA"
+  end
+end
```
justin808: Neat. Tests!

scripts/deploy
```diff
@@ -0,0 +1,39 @@
...
+echo git push $app $current_branch:master $@
+git push $app $current_branch:master $@
+
+echo Setting heroku config DEPLOYMENT_SHA
+echo heroku config:set DEPLOYMENT_SHA=$sha --remote $app
+heroku config:set DEPLOYMENT_SHA=$sha --remote $app
+
+# Run migrations always just in case
+heroku run rake db:migrate --remote $app
+heroku restart --remote $app
```
robwise: @dylangrafmyre Could you please help with implementing this deploy script somehow?

dylangrafmyre: @robwise @justin808 I really don't think we should be using this. Better just better to setup CI to deploy after build passes. Need to eliminate as may manual steps as possible.

robwise: @dylangrafmyre Is there a way to somehow set the `DEPLOYMENT_SHA` environment variable on Heroku to the latest commit on deployment automatically? That's all we're really trying to do.

robwise: @dylangrafmyre @justin808 What about instead of the `scripts/deploy` approach, we capture the `SOURCE_VERSION` environment variable ([doc](https://devcenter.heroku.com/changelog-items/630)) on Heroku? Justin, from the thread I found where you were looking into this before, it sounds like the variable is only available during the compilation phase of building the app. 

Therefore, our code to grab it would need to go in some sort of buildpack. Modifying the buildpack is probably a bad idea, but we can add our own custom buildpack that does ONLY this special thing. Heroku is apparently capable of running multiple buildpacks sequentially for the same build.

Actually, it looks like we are already running multiple buildpacks because of the need for both node and ruby. And this guy's project seems to be doing exactly what I was saying above: https://github.com/sreid/heroku-buildpack-sourceversion

I can make the necessary changes very easily if this is the route we want to go.

robwise:
Ready for review!

justin808:
LGTM! Great job!! Love the clean code and clean tests!

justin808:
Very nice!

PR #126, opened 2015-10-21
Refactor Feature Tests and Increase Coverage

robwise:


robwise:
I just thought they might be less confusing this way because before there were examples hidden inside of `shared_context`s that a reader would have no idea were in there based on the name. They also hurt reusability since the examples make assumptions about what you want to test instead of just focusing on the context in which you want to do it. Therefore, I made it so that shared examples may call shared contexts, but not the other way around.

I also split the different features into different files so that it would be easier find the specific feature tests you wanted to look at.

In the process, I noticed that some of the failure conditions were not being covered, so I wrote tests for them and realized that they were being covered. Therefore I added validation logic to pass those tests. While this is additional application logic, I think it's important that users are aware that they cannot rely on JavaScript front-end form validations only.

spec/features/destroy_comment_spec.rb
```diff
@@ -0,0 +1,16 @@
+require "rails_helper"
+require "features/shared/contexts"
+
+feature "Destroy a comment" do
+  include_context "Existing Comment"
+  context "from classic page" do
```
justin808: This makes we want to add a delete feature on the react side! :+1:

robwise: @justin808 Hey, don't look at me that was already in there!

justin808:
I like this. It takes a bit of getting used to the shared_context and include_context syntax. I've often used shared_examples, and now I'm going to consider using shared_context.

robwise:
@justin808 It definitely does take some getting used to as well as understanding how their DSL is working under the hood. This is why Minitest appeals to some developers because it just uses plain Ruby methods instead of making its own DSL so WYSIWYG (yay acronyms!). Sometimes I just write my own plain methods even in RSpec for this reason. But, the tests were already using `shared_context` and `shared_examples` before I refactored, so I just changed them around a little.

josiasds:
@robwise This is good!

justin808:
Thanks @robwise! LGTM!

PR #128, opened 2015-10-21
React Router integration

alexfedoseev:
**NB!** This branch will work only with these changes in gem: https://github.com/shakacode/react_on_rails/pull/68

When changes will be applied to `react_on_rails`, React Router will be working with gem on the server and on the client.

But still there are 2 things, that needs to be handled by the gem on the Rails side:

1. Errors from router. We should send 500 in case of the errors.
1. Redirects from router. We should send 302 in case of the redirect.

@justin808 @samnang Please take a look at [these TODOs](https://github.com/shakacode/react-webpack-rails-tutorial/blob/alex/react-router/client/app/startup/ServerApp.jsx#L11). And comment on 2 statements above.


client/package.json
```diff
@@ -49,9 +50,10 @@
     "loader-utils": "^0.2.11",
     "marked": "^0.3.5",
     "react": "^0.14.0",
-    "react-bootstrap": "^0.27.0",
+    "react-bootstrap": "https://registry.npmjs.org/react-bootstrap/-/react-bootstrap-0.27.2.tgz",
```
justin808: Unreleased?

client/app/startup/ServerApp.jsx
```diff
@@ -1,17 +1,38 @@
...
+  let routeProps;
+
+  match({ routes, location }, (_routeError, _routeRedirect, _routeProps) => {
+    routeError = _routeError;
+    routeRedirect = _routeRedirect;
+    routeProps = _routeProps;
+  });
+
+  if (routeError) {
+    // TODO: We should handle errors from router inside `react_on_rails` gem
```
justin808: @alexfedoseev Please open an issue on this for react_on_rails, and let's do it.

client/package.json
alexfedoseev: No, this issue: https://github.com/react-bootstrap/react-bootstrap/issues/1395
I can't do `npm shrinkwrap` with new version of `history` because of npm conflict with `react-bootstrap` -> `history` dependency.

client/app/startup/ServerApp.jsx
alexfedoseev: https://github.com/shakacode/react_on_rails/issues/69

samnang:
@alexfedoseev should the redirect still in context of react-router? or it needs to be client(browser) redirect? 

@justin808 As I remember we already catch any exception when we render component, right?

justin808:
@samnang Yes, we catch all errors and log them.
@alexfedoseev Can we simulate a likely error from react-router?

Maybe we just need to remove the TODO since we do catch and log errors already.

alexfedoseev:
> should the redirect still in context of react-router? or it needs to be client(browser) redirect?

We need to send server response with 302 redirect from Rails to new location (it will be returned by generator function).

> As I remember we already catch any exception when we render component, right?

Yep, it'll be logged in Development, but in Production we should send 500 response in case of server error, don't we?

@justin808 @samnang

justin808:
Maybe  react_on_rails should be configured to either throw or render errors when server rendering?

alexfedoseev:
I think it's more harm than good to give an option to treat server error as success response.

samnang:
I think we need to choose either throw or render errors, we don't have to support both and have an option. Too many options could be confusing.

I vote for throwing an exception with stack trace. It's easier for development.

> We need to send server response with 302 redirect from Rails to new location (it will be returned by generator function).

@alexfedoseev I'm a bit confused here, because after we render the component on server for the first request to the server, then after that everything will navigate at client side which depend on react-router to render different components, so why do we need to be done at server in this context? /cc @justin808

justin808:
Sure. We can change:

1. Throw exception if we catch a JS exception.
2. Handle the routing request.
3. Note, we already return the results back as a JSON array, so it's easy to add a 3rd element.

alexfedoseev:
@samnang Here is the use case:

Guest navigates to Profile Edit page. Guest can't access to this area and, before render the component, we check his status in transition hook. Inside the hook we declare that if user is not authorized -> redirect him to Login path. And the only thing that we have from the `match` function of React Router is info about redirect location. So we can't render anything, we have to send 302 redirect from server to the new location.

justin808:
@alexfedoseev We need to create some simple example in the tutorial, as well as something simple in the dummy apps to test against. I think real code will make this clear.

alexfedoseev:
@justin808 @samnang Here it is: c23b58a

Here we have pre-check: [method](https://github.com/shakacode/react-webpack-rails-tutorial/commit/c23b58a719f1bf0be1176b726fb8dc52f4becf21#diff-588b399fe7a56eda6f347ae9d245a356R5) + [hook](https://github.com/shakacode/react-webpack-rails-tutorial/commit/c23b58a719f1bf0be1176b726fb8dc52f4becf21#diff-a75ee0f35c15a87b2ed0e7c411799972R21)

If user will try to navigate to path with such pre-check:
* by clicking on link from client: he will be redirected by methods of History API
* by sending GET request to the server: router will tell us "Hey guys! I checked some stuff and looks like we have redirect here. Here is the info about redirect location, please, handle it. Bye-bye <3!"

So next we need to handle such cases on the Rails server.

alexfedoseev:
We discussed redirect thing with @justin808. Not sure how to return 302 redirect here, because we're in the middle of render cycle and we can't send 302 redirect from this point, because it's controller's thing. So let's stick with JS redirect for now. If someone has idea how to handle such case — let us know.

justin808:
@alexfedoseev We'll generated some JS in the rendering that sets window.location just like we generate JS to put server console messages.

justin808:
@robwise Can you give this change a try. I'll give you a hand.

alexfedoseev:
Do we want to make it in https://github.com/shakacode/react_on_rails/pull/68 or merge https://github.com/shakacode/react_on_rails/pull/68 and start new PR?

justin808:
@alexfedoseev Your call. It seems backwards compatible. It would be nice get some basic docs. @robwise can commit to this branch, so we don't need to start a new one. I want to get this in for version 1.0.0. Maybe we should release rc-pre-2?

robwise:
@justin808 @alexfedoseev I agree, committing to this would make more sense as really the tests and generators should go along with the code they relate to.

alexfedoseev:
:+1: @robwise Let's call tomorrow and discuss next steps in this PR.

robwise:
Just talked to Alex briefly on Slack, I should be able to get the generators branch done today or tomorrow at latest and then I can incorporate this into a generator as well as add some integration tests.

samnang:
@alexfedoseev Thanks for giving the scenario, but I'm a bit unclear because I feel all authorization should be on the server. If users try to navigate by changing browser url to unauthorized page, then Rails should check and redirect the request even they render the page. If the user know JS and try to change the rendering component or path in client only to access unauthorized sections/screens, we don't have to prevent it because even they see it, but they can't update it because all update requests will go to server and server will reject it. Those users are just attackers, so we don't treat them like normal flow. 

@justin808 @robwise

justin808:
I think what @samnang is referring to is why would the code even get to server rendering if the server knows the person is not allowed.

alexfedoseev:
> I feel all authorization should be on the server.

Source of truth — yes. All — not always. Because in case of SPA we need auth mechanism on the client to render UI and send auth headers with requests to API. And it's really simple example. There can be more complex cases. In SPA there is a form with 3 sections. User can't access 2nd section until he's done with 1st. So we need to add hook to 2nd section with logic: if 1st section is ok -> render 2nd, else -> redirect to 1st. And it's really annoying to duplicate routes + render logic on the server and on the client (just to send redirect), especially when SPA has only one entry point in the server routes.

samnang:
After had a chance to discuss with @alexfedoseev, It makes sense in a context of SPA.

justin808:
@mamute -- you could pull this branch and have it refer to the branch on react_on_rails with the fix. This can inspire your simple test.

https://github.com/shakacode/react_on_rails/pull/68

client/app/layout/Layout.jsx
```diff
@@ -0,0 +1,30 @@
...
+    children: PropTypes.object.isRequired,
+  }
+
+  render() {
+    return (
+      <section>
+        <header>
+          <ul>
+            <li>
+              <Link to="/">Comments</Link>
```
justin808: These will not automatically highlight. 

We need set the activeClassName property to true on the current link.

For an example of how to do this, react-router docs.

client/app/layout/Layout.jsx
```diff
@@ -0,0 +1,30 @@
+import React, { PropTypes } from 'react';
+import { Link } from 'react-router';
+
+export default class Layout extends React.Component {
+
+  static propTypes = {
+    children: PropTypes.object.isRequired,
```
justin808: This is where the content is provided for rendering by react router.

client/app/routes/routes.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import { Route } from 'react-router';
+import Layout from '../layout/Layout';
+import TestReactRouter from '../components/TestReactRouter';
+import TestReactRouterRedirect from '../components/TestReactRouterRedirect';
+import CommentScreen from '../components/CommentScreen';
+
+export default (
+  <Route component={Layout}>
+    <Route
```
justin808: Whatever renders here will go into the this.props.children for the Layout component!

client/app/routes/routes.jsx
```diff
@@ -0,0 +1,24 @@
...
+      component={CommentScreen}
+    />
+    <Route
+      path="/react-router"
+      component={TestReactRouter}
+    />
+    <Route
+      path="/react-router/redirect"
+      component={TestReactRouterRedirect}
+      onEnter={TestReactRouterRedirect.checkAuth}
```
justin808: This is where we're check server rendering doing the right thing for a redirect.

client/app/startup/ServerApp.jsx
```diff
@@ -1,17 +1,47 @@
 import React from 'react';
 import { Provider } from 'react-redux';
+import { match, RoutingContext } from 'react-router';
 
 import createStore from '../stores/commentsStore';
-import CommentScreen from '../components/CommentScreen';
+import routes from '../routes/routes';
 
-const App = props => {
+const App = (props, location) => {
```
justin808: New API for the function to take a location for server rendering.

client/app/startup/ServerApp.jsx
```diff
@@ -1,17 +1,47 @@
...
   const store = createStore(props);
-  const reactComponent = (
+
+  let routeError;
+  let routeRedirect;
+  let routeProps;
+
+  match({ routes, location }, (_routeError, _routeRedirect, _routeProps) => {
```
justin808: This run synchronously because we are not setting up the app with dynamic routes.

This does not affect the react_on_rails gem because it does not support dynamic routes.

@robwise FYI -- note for you in the docs.

client/app/startup/ServerApp.jsx
```diff
@@ -1,17 +1,47 @@
...
+    routeError = _routeError;
+    routeRedirect = _routeRedirect;
+    routeProps = _routeProps;
+  });
+
+  if (routeError) {
+    // TODO: We should handle errors from router inside `react_on_rails` gem
+    //       routeError: { message: "Error message <String>" }
+    throw new Error(routeError.message);
```
justin808: We've never tested this. We don't know what causes this.

client/app/components/TestReactRouterRedirect.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+export default class TestReactRouterRedirect extends React.Component {
+
+  static checkAuth(nextState, replaceState) {
+    const notAuthorized = true;
+    if (notAuthorized) {
```
justin808: intentional to show the logic....

client/app/components/TestReactRouterRedirect.jsx
```diff
@@ -0,0 +1,18 @@
...
+      replaceState({ redirectFrom: nextState.location.pathname }, '/');
+    }
+  }
+
+  render() {
+    return (
+      <div>Nope.</div>
```
justin808: this rendering never actually happens b/c we hard code checkAuth to redirect

client/app/components/CommentScreen.jsx
```diff
@@ -15,13 +15,29 @@ const CommentScreen = React.createClass({
...
+    );
   },
 
   render() {
     const { dispatch, data } = this.props;
     const actions = bindActionCreators(commentsActionCreators, dispatch);
+
     return (
       <div>
+        {::this._renderNotification()}
```
justin808: TODO: convert binding to constructor.

client/app/components/CommentScreen.jsx
```diff
@@ -15,13 +15,29 @@ const CommentScreen = React.createClass({
   propTypes: {
     dispatch: PropTypes.func.isRequired,
     data: PropTypes.object.isRequired,
+    location: PropTypes.shape({
+      state: PropTypes.object,
+    }).isRequired,
```
justin808: The location prop type comes from ReactRouter!

client/app/routes/routes.jsx
```diff
@@ -0,0 +1,24 @@
...
+import Layout from '../layout/Layout';
+import TestReactRouter from '../components/TestReactRouter';
+import TestReactRouterRedirect from '../components/TestReactRouterRedirect';
+import CommentScreen from '../components/CommentScreen';
+
+export default (
+  <Route component={Layout}>
+    <Route
+      path="/"
+      component={CommentScreen}
```
justin808: This forward all props and adds the prop "location"

client/app/startup/ClientApp.jsx
```diff
@@ -1,17 +1,20 @@
...
 
 const App = props => {
   const store = createStore(props);
-  const reactComponent = (
+  const history = createHistory();
+
+  return (
     <Provider store={store}>
-      <CommentScreen />
+      <Router history={history} children={routes} />
```
justin808: if we were not using redux, then we'd provide to the router some kind of middleware, with a prop called "createElement", which would bind the props with a component: https://github.com/rackt/react-router/blob/master/docs/API.md#createelementcomponent-props

```
<Router createElement={createElement} />

// default behavior
function createElement(Component, props) {
  // make sure you pass all the props in!
  return <Component {...props}/>
}
```

client/app/components/TestReactRouterRedirect.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+export default class TestReactRouterRedirect extends React.Component {
+
+  static checkAuth(nextState, replaceState) {
+    const notAuthorized = true;
+    if (notAuthorized) {
+      replaceState({ redirectFrom: nextState.location.pathname }, '/');
```
justin808: This puts the object in the "state" of the routerRedirect object that the callback will get

    //       routeRedirect: {
    //         pathname: "Path <String>",
    //         search: "Query <String>",
    //         hash: "Hash <string>",
    //         state: "Custom State <Object>",
    //         ...
    //        }

justin808:
This is definitely ready. See https://github.com/shakacode/react_on_rails/pull/68.

You can pull this branch and the gemfile refers to the branch for react_on_rails.

justin808:
@alexfedoseev @robwise @jbhatab @ryanaip @rstudner @thiagoc7, please review.

alexfedoseev:
LGTM :+1:

justin808:
@alexfedoseev Please see last commit https://github.com/shakacode/react-webpack-rails-tutorial/commit/f279c5721e60625479e60584f9e2b2fd7fd905f8 and let use know what you think.

@robwise Please be sure to update the Gemfile before merging. Actually, once we release, we put the version of the gem at 1.1.0.

justin808:
@robwise you've got some test failures after your last push.

alexfedoseev:
@justin808 @robwise Looks good! :+1:

robwise:
@justin808 Aren't those coming from yours (https://github.com/shakacode/react-webpack-rails-tutorial/commit/2feb2e7cb03b585218f71f25e378391b934d29b7)?

justin808:
Once @alexfedoseev gets a chance to :+1: on the refactoring of the RouterCommentScreen, NonRouterCommentScreen, CommentScreen components, we'll merge!

alexfedoseev:
All good! :+1:

josiasds:
It looks like I've missed all the action here. Great job! :+1: @justin808 @alexfedoseev @robwise

PR #129, opened 2015-10-22
Change calls of ReactOnRails to Rails.application

robwise:
Ensure that users copying example files won't run into issues due to references to `RailsReactTutorial::Application`, which obviously won't be the name of their app. Instead, use `Rails.application`. This reflects similar changes in https://github.com/shakacode/react_on_rails/pull/53

Besides the actual `config/application.rb` file and the `<title>` element of the main layout, I found no other occurrences.

### DIFF
--- a/lib/tasks/assets.rake
+++ b/lib/tasks/assets.rake
@@ -18,7 +18,7 @@ namespace :assets do
   end
 
   task :clobber do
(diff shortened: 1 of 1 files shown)


justin808:
Thanks @robwise! LGTM!

PR #131, opened 2015-10-29
Update application.js

justin808:


robwise:
Just as a note: we determined this was not necessary because the shim is already being included in the webpack, so we don't need to also include it in sprockets.

hisapy:
Based on this merge, would it be ok to remove **gem 'es5-shim-rails'** from Gemfile?

robwise:
@hisapy Yes, good point, I'm making a new PR to incorporate that. https://github.com/shakacode/react-webpack-rails-tutorial/pull/132

hisapy:
Thx for the quick response @robwise

PR #132, opened 2015-10-29
Remove es5-shim from Gemfile

robwise:
No longer necessary as we are relying on Webpack to use its own es5-shim loader.

Thanks to @hisapy: https://github.com/shakacode/react-webpack-rails-tutorial/pull/131#issuecomment-152184021


robwise:
@justin808 Based on our discussion last night, I'm pretty sure this is correct, but could you confirm?

justin808:
LGTM.

PR #136, opened 2015-10-30
updates for readability

justin808:
Also, add comment to database.yml

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,9 +4,15 @@ ruby "2.2.2"
 #
 # Bundle edge Rails instead: gem "rails", github: "rails/rails"
 gem "rails", "~>4.2"
-# Use sqlite3 as the database for Active Record
-gem "sqlite3", group: :development
-gem "pg", group: :production
+
+# Note: We're using sqllite3 for development and testing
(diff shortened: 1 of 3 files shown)


justin808:
@robwise Check this out.

Gemfile
```diff
@@ -4,9 +4,15 @@ ruby "2.2.2"
 #
 # Bundle edge Rails instead: gem "rails", github: "rails/rails"
 gem "rails", "~>4.2"
-# Use sqlite3 as the database for Active Record
-gem "sqlite3", group: :development
-gem "pg", group: :production
+
+# Note: We're using sqllite3 for development and testing
+gem "sqlite3", group: [:development, :test]
```
robwise: Why are we using group blocks for some gems (see line 55, for example) and passing them as hash options for others?

samnang: I think maybe no reasons for this, but sometimes the order of loading gems matter.

justin808: Whatever makes sense for readability. Usually the DB gems go at the top.

samnang: Totally agree with @justin808

Issue #139, opened 2015-10-30
Switch to ES6 Class syntax for React from ES5 React.createClass


justin808:
@alexfedoseev will comment some more on the reasons we should do this.

Any volunteers for converting the code?

@robwise We'll be updating the generator to support this.

rstudner:
I'll start on it tonight. Outside of cases using lonkedstatemixin, I've been doing es6 only react stuff and am familiar with he converting process.

robwise:
@justin808 @rstudner this is implemented in https://github.com/shakacode/react_on_rails/pull/70 if you want to get an idea for what the syntax looks like. Note also that we are using a new paradigm in terms of `client/` folder organization.

For more on that, see this doc (soon to be included in the gem README): https://docs.google.com/a/friendsandguests.com/document/d/1jBrIjOfZ3ct8Uo2mzzwO0wCCkLtcl-fXyhgIFMKQitY/edit?usp=sharing

rstudner:
It seems that the code is using es6 function syntax..  ala -- render() {  }  but is still using the React.createClass ES5 syntax (for say ,the helloworldcontainer).

Sorry, on mobile, hard to do markdown :)

rstudner:
Thanks @robwise.  I think I have all the code updated above (i'd be a bit too aggressive in removing commas ;).  

I was having problems with the :: es7 binding syntax, so I used the ES6+/ES6 fat arrow function syntax.

I can revise it again if people really prefer it over the fat arrow (I guess I sort of like the fat arrow because it serves the same role as how coffeescript uses the fat arrow and more people might be familiar with it).

Either way is cool with me :)

robwise:
@rstudner Yeah I didn't get the `::this` syntax until @jbhatab showed me it was exactly the same as doing `this.someMethod().bind(this)`. Basically, class methods by default have their `this` reference the class and not the actual component instance, so when we call methods from `render` (at which point `this` would refer to the component) we want to pass that binding so that the methods can access that component's `refs` and such.

I also wrote a couple of comments as I think it's considered better to not use the constructor if you don't need it. @alexfedoseev @jbhatab Can you comment on this?

[...]

PR #140, opened 2015-10-31
ES6 all React stuff.

rstudner:
I also cleaned up all the trailing commas (that where everywhere) because they are lint etc issues (rubymine in particular really dislikes them). They hurt nothing to take out.

@justin808 After submitting a comment, which works, there is still an issue (that seems like) a "this.props" where this isn't defined.

You'll notice in the PR various uses of .bind(this).. that is because ES6 react classes don't automagically do that anymore (so like, a clickHandler method cannot refer to "this" (which they all do) unless you do this._handleClick.bind(this) sort of syntax)

Clearly I missed one/haven't unraveled it yet :)  That being said, this has to be 95-99% there and this way more eyes can get on it.

### DIFF
--- a/client/app/components/Comment.jsx
+++ b/client/app/components/Comment.jsx
@@ -1,13 +1,12 @@
 import React, { PropTypes } from 'react';
 import marked from 'marked';
 
-const Comment = React.createClass({
-  displayName: 'Comment',
-
-  propTypes: {
+class Comment extends React.Component {
+  static displayName = 'Comment';
+  static propTypes = {
     author: PropTypes.string.isRequired,
     text: PropTypes.string.isRequired,
-  },
+  };
 
   render() {
     const { author, text } = this.props;
@@ -20,7 +19,7 @@ const Comment = React.createClass({
         <span dangerouslySetInnerHTML={{__html: rawMarkup}}/>
       </div>
     );
-  },
-});
+  }
+}
 
 export default Comment;
--- a/client/app/components/CommentBox.jsx
+++ b/client/app/components/CommentBox.jsx
@@ -8,8 +8,7 @@ class CommentBox extends React.Component {
     super(props, context);
     this.state = {};
 
-    this._ajaxCounter = this._ajaxCounter.bind(this);
-    this._isSendingAjax = this._isSendingAjax.bind(this);
+    _.bindAll(this, '_ajaxCounter', '_isSendingAjax');
   }
 
   static displayName = 'CommentBox';
--- a/client/app/components/CommentBox.jsx
+++ b/client/app/components/CommentBox.jsx
@@ -4,6 +4,14 @@ import CommentForm from './CommentForm';
 import CommentList from './CommentList';
 
 class CommentBox extends React.Component {
+  constructor(props, context) {
+    super(props, context);
+    this.state = {};
+
+    this._ajaxCounter = this._ajaxCounter.bind(this);
+    this._isSendingAjax = this._isSendingAjax.bind(this);
+  }
+
   static displayName = 'CommentBox';
   static propTypes = {
     pollInterval: PropTypes.number.isRequired,
@@ -17,12 +25,12 @@ class CommentBox extends React.Component {
     setInterval(fetchComments, this.props.pollInterval);
   }
 
-  ajaxCounter() {
+  _ajaxCounter() {
     return this.props.data.get('ajaxCounter');
   }
 
-  isSendingAjax() {
-    return this.ajaxCounter() > 0;
+  _isSendingAjax() {
+    return this._ajaxCounter() > 0;
   }
 
   render() {
@@ -31,14 +39,14 @@ class CommentBox extends React.Component {
     return (
       <div className="commentBox container">
         <h2>
-          Comments { this.isSendingAjax() && `SENDING AJAX REQUEST! Ajax Counter is ${this.ajaxCounter()}` }
+          Comments { this._isSendingAjax() && `SENDING AJAX REQUEST! Ajax Counter is ${this._ajaxCounter()}` }
         </h2>
         <p>
           Text take Github Flavored Markdown. Comments older than 24 hours are deleted.
           <b>Name</b> is preserved, <b>Text</b> is reset, between submits.
         </p>
         <CommentForm
-          ajaxSending={this.isSendingAjax()}
+          ajaxSending={this._isSendingAjax()}
           error={data.get('submitCommentError')}
           actions={actions}
         />
--- a/client/app/components/CommentBox.jsx
+++ b/client/app/components/CommentBox.jsx
@@ -3,28 +3,27 @@ import React, { PropTypes } from 'react';
 import CommentForm from './CommentForm';
 import CommentList from './CommentList';
 
-const CommentBox = React.createClass({
-  displayName: 'CommentBox',
-
-  propTypes: {
+class CommentBox extends React.Component {
+  static displayName = 'CommentBox';
+  static propTypes = {
     pollInterval: PropTypes.number.isRequired,
     actions: PropTypes.object.isRequired,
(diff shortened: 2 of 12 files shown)


justin808:
@rstudner Thanks so much! ES6 is great. However, we're trying to keep with the AirBnb JS style guideline as much as possible. The trailing commas are there for a reason.

For the bind parts, we want to use the ES7 `::` syntax.

@alexfedoseev anything else?

justin808:
Connects to #139

rstudner:
Ahh!  my apologies on their removal then! :)  I can go put them back in easily enough.

Sorry, I didn't think to 'go ES7' (despite of course the Readme.MD saying ES7 ;).  I was mostly just trying to mimic syntax people would see if they were reading the official React docs and weren't familiar with ES7.

client/app/components/Comment.jsx
```diff
@@ -1,13 +1,16 @@
 import React, { PropTypes } from 'react';
 import marked from 'marked';
 
-const Comment = React.createClass({
-  displayName: 'Comment',
-
-  propTypes: {
+class Comment extends React.Component {
+  constructor(props) {
```
robwise: I think it's better to avoid using `constructor` in React components, especially since we aren't even setting `state` here. If we do need to set state, it should be via class a class property. For example, if we want to have a state of `{author: "Timmy", text: "hello world"}, you just do this:

```javascript
class Comment extends React.Component {
  author = "Timmy";
  text = "hello world";
....
```

client/app/components/CommentBox.jsx
```diff
@@ -3,28 +3,32 @@ import React, { PropTypes } from 'react';
 import CommentForm from './CommentForm';
 import CommentList from './CommentList';
 
-const CommentBox = React.createClass({
-  displayName: 'CommentBox',
-
-  propTypes: {
+class CommentBox extends React.Component {
+  constructor(props) {
```
robwise: see https://github.com/shakacode/react-webpack-rails-tutorial/pull/140/files#r43574609

client/app/components/CommentForm.jsx
```diff
@@ -10,31 +10,30 @@ import ReactCSSTransitionGroup from 'react/lib/ReactCSSTransitionGroup';
 const emptyComment = { author: '', text: '' };
 const textPlaceholder = 'Say something using markdown...';
 
-const CommentForm = React.createClass({
-  displayName: 'CommentForm',
+class CommentForm extends React.Component {
+  constructor(props) {
```
robwise: see https://github.com/shakacode/react-webpack-rails-tutorial/pull/140/files#r43574609

client/app/components/Comment.jsx
```diff
@@ -1,13 +1,12 @@
 import React, { PropTypes } from 'react';
 import marked from 'marked';
 
-const Comment = React.createClass({
-  displayName: 'Comment',
-
-  propTypes: {
+class Comment extends React.Component {
+  static displayName = 'Comment';
```
alexfedoseev: It can be safely removed in ES6 classes syntax.

client/app/components/CommentForm.jsx
```diff
@@ -10,31 +10,27 @@ import ReactCSSTransitionGroup from 'react/lib/ReactCSSTransitionGroup';
...
-  getInitialState() {
-    return {
-      formMode: 0,
-      comment: emptyComment,
-    };
-  },
+  };
 
-  handleSelect(selectedKey) {
+  handleSelect = (selectedKey) => {
```
yorzi: @rstudner I am new to React/ES6, I am wondering why is this pattern better that the original one?

justin808: @yorzi , @alexfedoseev might have a good reference.

justin808:
@rstudner the only CI failure is linting.

@alexfedoseev maybe comment if we should go with this?

alexfedoseev:
@justin808 @rstudner This commit: https://github.com/shakacode/react-editable-grid-example/commit/f7968e1b562608569b6c2f160642a64d8ed8c080

There are 2 branches:

* [master](https://github.com/shakacode/react-editable-grid-example/tree/master) (with functions, working)
* [fat-arrows](https://github.com/shakacode/react-editable-grid-example/tree/fat-arrows) (with fat-arrows functions, broken)

Please confirm, that second one is broken.

rstudner:
Pushed an update to make the lint-robot happy :)

In regards to fat arrow vs :: and other syntax things, the longest(?) explanation I found for them was here:
http://egorsmirnov.me/2015/08/16/react-and-es6-part3.html

alexfedoseev:
The rule is simple: fat arrows works without binding only within one component. And it's broken when they are passed to outer world. When fat arrow functions are used, we have to keep track of the cases when we have to bind function to context and when it's not necessary. So the question is: do we need one more thing to remember?

alexfedoseev:
All in all, I vote for binding methods to context in constructor. [One more reason](http://alexfedoseev.com/post/65/react-event-handlers-and-context-binding), btw.

rstudner:
@alexfedoseev - I 100% agree with binding methods in the constructor.  Earlier in this (long hah) PR of comments I was told to not use constructors.

Thus -- I went with what I thought was the 'next best' option.

As well, in a Redux world I assumed we wouldn't be doing the 'passing of callback functions down via props' and thus none of these would leak outside of their component of origination.

alexfedoseev:
> As well, in a Redux world I assumed we wouldn't be doing the 'passing of callback functions down via props' and thus none of these would leak outside of their component of origination.

It happens quite often, for example when we need to pass the callback to save / close reusable modal dialog.

rstudner:
@alexfedoseev Ahh, i've been handling it differently just on my own personal projects -- but yeah, that would be a widely encountered use case for sure.

Based on your own blog post, which syntax would you go for (you are using .bind(this) in the constructors and as well, earlier in this PR, I was told not to use .bind(this)).

alexfedoseev:
Let's stick with:

```js
constructor(props, context) {
  super(props, context);

  this._handleSomething = this._handleSomething.bind(this);
}
```

It's less problematic way of doing binding.

We can do something like:

```js
// bindToContext.js
export default methods => {
  methods.forEach(method => this[method] = this[method].bind(this))
}

// Component
constructor(props, context) {
  super(props, context);

  bindToContext([ '_handleSomething' ]).bind(this);
}
```

rstudner:
Had a good quick chat with Alex over Slack.  Will make some updates in a couple hours (have to feed small childrenBeasts and get them on the bus)

justin808:
@rstudner Let's squash and create a nice commit message that sums up the choices. And then we'll merge.

client/app/components/CommentBox.jsx
```diff
@@ -3,28 +3,35 @@ import React, { PropTypes } from 'react';
...
-const CommentBox = React.createClass({
-  displayName: 'CommentBox',
+class CommentBox extends React.Component {
+  constructor(props, context) {
+    super(props, context);
+    this.state = {};
 
-  propTypes: {
+    this._ajaxCounter = this._ajaxCounter.bind(this);
+    this._isSendingAjax = this._isSendingAjax.bind(this);
```
justin808: one last change now that we're using lodash:

```
_.bindAll(this, '_ajaxCounter', '_isSendingAjax');
```

rstudner:
Thoughts on the above?  I can import lodash in each file, or we can expose it.  the above commit just exposes it like react/jquery.

alexfedoseev:
We should expose something only for the reason. For example, `React` should be exposed for the gem or `fetch` should be exposed since it's (probably) future global object. There is no reason to expose `lodash`, so it should be imported.

rstudner:
Okay, i'll change to imports.  Interestingly, exposing it solved 99% of the errors, but not all of them (which is odd all by itself)

PR #145, opened 2015-11-02
Trying out all the assets configuration diff spot

justin808:
# Add client/assets/stylesheets to asset pipeline's search path.
    config.assets.paths << Rails.root.join("client", "assets", "stylesheets")


samnang:
:+1:

Issue #124, opened 2015-10-16
Consider including Lodash in the project


justin808:
Take a look at this:
https://github.com/shakacode/react-webpack-rails-tutorial/pull/123/files#r42269891

The take away is this:

```javascript
Array.from(document.getElementsByTagName('meta')).find(tag => tag.name === 'csrf-token')
```

vs with Lodash:

```javascript
_.find(document.querySelectorAll('meta'), 'name', 'csrf-token')
```

I find lodash to be my swiss army knife when coding JS, and it fits in well with React.

In fact, Lodash is already an implicit dependency.

https://github.com/shakacode/react-webpack-rails-tutorial/blob/master/client/npm-shrinkwrap.json#L2669

Anybody second me?

@alexfedoseev @samnang @josiasds @mapreal19 @dylangrafmyre @rstudner?

samnang:
:+1:

justin808:
https://www.youtube.com/watch?v=Pq1RD22W7B4#t=55s

I vote for lodash.

CC: @jdalton

justin808:
More on lodash: https://github.com/Khan/style-guides/pull/28/files#files

My personal preference is to lean on lodash for consistency. It's a lot of work to keep track of both, and like using the linters, I prefer consistency.

justin808:
@josiasds I put your name on this one. Should be quick. 

@alexfedoseev if we put lodash in the vendor bundle, I'm pretty sure that parts we don't user don't get stripped out. That's a need for jQuery. I'm thinking we could have a generated/vendor-optimized.js bundle that has all the other libraries we include with unused stripped out. 

I'd like to be able to see the project, non-vendor, code in it's own generated file, as that's useful for explaining what Webpack does.

robwise:
I know it's nice to have the webpage download as small as possible, but how many KB are we talking about here? If we use half of the lodash library and leave the other half unused, we'd be foregoing, what, 9KB of savings by just importing all of it?

justin808:
Fixed in #147

PR #147, opened 2015-11-03
Add lodash and update dependencies

josiasds:
Add lodash as discussed in #124.


client/app/utils/metaTagsManager.js
```diff
@@ -1,3 +1,5 @@
+import _find from 'lodash/collection/find';
```
josiasds: According to the [lodash docs](https://lodash.com/#installation), loading only the required method is "great for smaller builds with browserify/webpack". I kept the `_` in the method name just for consistency with the library.

justin808:
LGTM

PR #151, opened 2015-11-06
mocha tests and update dependencies

justin808:
This PR adds a JS unit testing framework with mocha and chai.
Sample unit tests are provided for Comment and CommentList.

Also updates react to 0.14.1 to satisfy a test-utils dependency


justin808:
@glennr Here's the update. If this passes, we merge.

@robwise: We can then add this to the generators.

justin808:
@dylangrafmyre @robwise @alexfedoseev @glennr I just noticed that we're not running these tests from our main CI script. What's the best way to do that?

PR #155, opened 2015-11-16
enable react animations for comments list

thiagoc7:
Just following the [guide](https://facebook.github.io/react/docs/animation.html), and using the same `element` class that already exist.

### DIFF
--- a/client/app/components/CommentList.jsx
+++ b/client/app/components/CommentList.jsx
@@ -39,7 +39,7 @@ class CommentList extends React.Component {
       // http://facebook.github.io/react/docs/multiple-components.html#dynamic-children
       return (
         <Comment
-          key={index}
+          key={$$comment.get('id')}
           author={$$comment.get('author')}
           text={$$comment.get('text')}
         />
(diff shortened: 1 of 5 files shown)


rstudner:
I believe ReactCSSTransitionGroups will throw console warnings still.  It was discovered in 0.14.0, and sadly not yet fixed in 0.14.2 (it will be in 0.14.3).

Do you see warnings in your console when you do the animations?

thiagoc7:
no warnings... checked on both 3000 and 4000 ports.
Maybe you had warnings due to using `index` as key, it just works with `unique id`.

I just don't know how to fix the issues on travis.

rstudner:
Huh.. interesting/good.

This is the issue I was referring to:
https://github.com/facebook/react/issues/4876

client/server.js
```diff
@@ -5,10 +5,11 @@ var WebpackDevServer = require('webpack-dev-server');
 var jade = require('jade');
 var sleep = require('sleep');
 var config = require('./webpack.client.hot.config');
+var uuid = require('node-uuid')
 
 var comments = [
-  { author: 'Pete Hunt', text: 'Hey there!' },
-  { author: 'Justin Gordon', text: 'Aloha from @railsonmaui' },
+  { author: 'Pete Hunt', text: 'Hey there!', id: uuid.v4() },
```
justin808: Why is this added?

client/app/components/CommentList.jsx
```diff
@@ -39,7 +39,7 @@ class CommentList extends React.Component {
       // http://facebook.github.io/react/docs/multiple-components.html#dynamic-children
       return (
         <Comment
-          key={index}
+          key={$$comment.get('id')}
```
justin808: Why is this change necessary?

justin808:
Hi @thiagoc7. Maybe rebase on top of master and let me know what the next step is.

client/server.js
thiagoc7: we need a unique id for list animations

client/app/components/CommentList.jsx
thiagoc7: to get the unique id, or animations won't work

justin808:
@thiagoc7 @josiasds @mamute, let's see about getting this merged.

josiasds:
Finished on #164.

PR #158, opened 2015-11-17
new loading state tracker

thiagoc7:
Use isSaving and isFetching instead of ajaxCount.

If you guys like this solution, we need to think a good way to set a `minimum delay` before setting `loading state` back to false.

If it's too fast, like on localhost, it's not a good UX.

Thanks.


justin808:
@alexfedoseev How does this look for you?

alexfedoseev:
Looks good! :+1: 

According to minimum delay. It can smth like this:

```js
const minimumDelay = 500;

function handleAddCommentSuccess(dispatch, comment) {
  dispatch({
    type: COMMENT_ADD_SUCCEED,
    comment,
  });
}

export function addComment(comment) {
  return dispatch => {
    const startTime = (new Date()).getTime();
    dispatch({
      type: COMMENT_ADD_REQUESTED,
    });

    apiCall({ params })
      .then(res => {
        const comment = res.data;
        const successTime = (new Date()).getTime();
        const timeDiff = successTime - startTime;
        if (timeDiff < minimumDelay) {
          setTimeout(
            handleAddCommentSuccess.bind(null, dispatch, comment),
            minimumDelay - timeDiff
          );
        } else {
          handleAddCommentSuccess(dispatch, comment);
        }
      })
      .catch(res => {
        // same here...
      });
  }
}
```

We can wrap it in a nice helper.

thiagoc7:
Thanks @alexfedoseev!

I just [changed](http://arguments.callee.info/2008/11/10/passing-arguments-to-settimeout-and-setinterval/) a a little the `setTimeout`, so we can avoid dispatch as an argument.

client/server.js
```diff
@@ -35,7 +34,6 @@ server.app.get('/comments.json', function(req, res) {
 server.app.post('/comments.json', function(req, res) {
   console.log('Processing comment: %j', req.body.comment);
   console.log('(shhhh...napping 1 seconds)');
-  sleep.sleep(1);
```
justin808: @thiagoc7 Why did you remove this?

client/app/reducers/commentsReducer.js
```diff
@@ -20,12 +21,14 @@ export default function commentsReducer($$state = $$initialState, action = null)
       return $$state.merge({
         $$comments: comments,
         fetchCommentError: null,
+        isFetching: false,
```
justin808: We also show an ajax indicator when saving the record.

If that is the case, why call the flag `isFetching`? A reference to the fetch API?

How about `ajaxBusy`?

client/app/actions/commentsActionCreators.js
```diff
@@ -43,24 +45,34 @@ export function submitCommentFailure(error) {
 
 export function fetchComments() {
   return dispatch => {
+    dispatch(setIsFetching());
     return (
-      commentsManager
-        .fetchComments()
-        .then(res => dispatch(fetchCommentsSuccess(res.data)))
-        .catch(res => dispatch(fetchCommentsFailure(res.data)))
+        commentsManager
```
justin808: why the extra indentation?

In general, try to limit whitespace diffs unless the linter is complaining.

Maybe the linter wants extra indentation?

client/app/actions/commentsActionCreators.js
```diff
@@ -43,24 +45,34 @@ export function submitCommentFailure(error) {
...
+          const result = res.data;
+          const successTime = (new Date()).getTime();
+          const timeDiff = successTime - startTime;
+          if (timeDiff < minimumDelay) {
+            setTimeout(dispatch, minimumDelay - timeDiff, submitCommentSuccess(result));
+          } else {
+            dispatch(submitCommentSuccess(result));
+          }
+        })
         .catch(res => dispatch(submitCommentFailure(res.data)))
```
justin808: I'm a bit confused on setting a minimum delay on the client.

I had put in a sleep on the server to simulate a somewhat slow connection.

justin808:
Hi @thiagoc7, I had a few questions for you. Thanks for contributing this.

client/server.js
thiagoc7: @justin808, I removed because we don't need this if we the minimum delay func.
Let's talk about this below.

client/app/reducers/commentsReducer.js
thiagoc7: ;)

client/app/actions/commentsActionCreators.js
thiagoc7: The idea of minimum delay is just for UX.
If the promise return very fast, you just have a blink, and it's kind of confusing.
But I agree that it's too much code for a starter repo...

thiagoc7:
Hey @justin808, do you want me to remove this `minimum delay` stuff?
I also think that it's much more clear the way it were, with the `sleep` at server file.

I'll try to create a middleware for this, just don't know how yet... I'll take a look at tutorials this weekend.

client/app/actions/commentsActionCreators.js
justin808: There's no need for middleware if we're just going put back the sleep in the server. The middleware was for the idea of having the action start a timeout for the processing indicator. I agree with you that it's just too much to get into having this extra code in the submit action. It's already complicated enough with the thunk middleware to do the asynch save.

client/app/actions/commentsActionCreators.js
thiagoc7: don't know what happened, it's fixed

justin808:
@alexfedoseev ready to merge?

alexfedoseev:
Looks good! :+1:

justin808:
@robwise Let's merge this.

PR #165, opened 2015-12-02
setting node version on travis

mamute:


.travis.yml
```diff
@@ -4,6 +4,8 @@ rvm:
   - 2.2.3
 install:
   - bundle install
+  - nvm install 5.0
+  - nvm use 5.0
   - npm install -g npm
```
dylangrafmyre: @justin808 @mamute Since we are now install Node 5.0 with nvm, we can probably remove `npm install -g npm`. Node 5.0 comes with NPM 3.x.x.

Before travis was using a older version of npm that was failing our build.

mamute: done :+1:

justin808: Hi @dylangrafmyre I just merged. Let's do another PR if that's the case.

BTW, I get the linting error locally, so this is probably not the main build issue. There's some problem with eslint?

dylangrafmyre: @justin808 the main build failure is from a c++ compiler error for `NAN` module used in sleep. I will fix this.

PR #167, opened 2015-12-03
remove unnecessary this binding for helper (ONLY needed for callbacks)

mamute:


client/app/components/RouterCommentScreen.jsx
```diff
@@ -10,6 +11,12 @@ function select(state) {
 }
 
 class RouterCommentScreen extends React.Component {
+  constructor(props, context) {
+    super(props, context);
+
+    _.bindAll(this, '_renderNotification');
```
justin808: @mamute Can you check this is actually required, meaning a binding of this.

The reason I ask is that it looks like this._renderNotifcation is simply a helper that returns either null or a react component.

I suspect that the fix is that that the `::` was not needed.

mamute:
@justin808 you're right, i've checked and fixed

justin808:
Thanks. @alexfedoseev FYI...

@mamute I updated the title.

mamute:
:+1:

PR #166, opened 2015-12-02
Fix ci build failures

dylangrafmyre:
@justin808, @robwise, @alexfedoseev  
I was able to fix the failing Travis CI build with setting Travis to install package `g++-4.9`.

However, I fixed the failing Codeship build by setting `react/prop-types: 0` within `.eslintrc`. 

This was the build fail issue:

1__dylan_dylans-macbook-pro____react-webpack-rails-tutorial_client__zsh_

And referenced here in the issues for `eslint-plugin-react`: 
https://github.com/yannickcr/eslint-plugin-react/issues/9

I am sure you will not want to have this rule continually ignored; therefore, someone will need to address the issue in the code?

Thanks,
Dylan

PR #168, opened 2015-12-03
Fix eslint plugin error

josiasds:
I did some debugging and find out the line that was failing the linter, so I just rewrote it.

### DIFF
--- a/client/app/stores/commentsStore.js
+++ b/client/app/stores/commentsStore.js
@@ -5,11 +5,10 @@ import reducers from '../reducers';
 import { initalStates } from '../reducers';
 
 export default props => {
-  const initialComments = props;
   const { $$commentsState } = initalStates;
   const initialState = {
     $$commentsStore: $$commentsState.merge({
-      $$comments: initialComments,
+      $$comments: props,
     }),
   };
 


josiasds:
Merged on #166.

client/app/stores/commentsStore.js
```diff
@@ -5,11 +5,10 @@ import reducers from '../reducers';
 import { initalStates } from '../reducers';
 
 export default props => {
-  const initialComments = props;
   const { $$commentsState } = initalStates;
   const initialState = {
     $$commentsStore: $$commentsState.merge({
-      $$comments: initialComments,
+      $$comments: props,
```
justin808: This is bizarre.

1. Please back out this change. The old code is important for readability.
2. we should ignore the eslint rule here and with a comment to the github issue indicating the problem.
3. please put a link to this code in the eslint issue (actually react-plugin probably, not main eslint)

Thanks @josiasds

justin808:
@josiasds @dylangrafmyre This needs to be reverted. See prior comments. Thanks.

client/app/stores/commentsStore.js
dylangrafmyre: @josiasds @justin808 Maybe you can just disable the rule for only this file or block? and not for the whole app. Don't know if the plugin will allow this?

josiasds:
Previous change reverted and eslint rule disabled on commit bc076958012b176fa5aee1f5347e5153400b2426.

client/app/stores/commentsStore.js
josiasds: @dylangrafmyre yep, we just need to put `/* eslint react/prop-types: 0 */` at the top of the file, see bc076958012b176fa5aee1f5347e5153400b2426.

josiasds: @justin808 Done :+1:

justin808: :+1:

PR #170, opened 2015-12-06
Update package eslint-plugin-react to fix bug

josiasds:
The package `eslint-plugin-react` was updated to fix this issue: https://github.com/yannickcr/eslint-plugin-react/issues/345#event-483515053

So now we can re-enable the eslint rule that was crashing our `commentStore`. (see #168 and #166).


justin808:
Thanks @josiasds !

josiasds:
Rebased on top of master.

Issue #141, opened 2015-10-31
Include pure-render-decorator


justin808:
If we're using Immutable.js, we should use the pure-render mixin. However, we're about to change to ES6 classes, so this seems to fit the bill.

```javascript
import {Component} from 'react';
import pureRender from 'pure-render-decorator';

@pureRender
class Test extends Component {
  render() {
    return <div></div>;
  }
}
```

* https://facebook.github.io/react/docs/pure-render-mixin.html
* https://github.com/felixgirault/pure-render-decorator
* https://www.npmjs.com/package/pure-render-decorator
* [React.js Conf 2015 - Immutable Data and React](https://www.youtube.com/watch?v=I7IdS-PbEgI)

CC: @alexfedoseev @rstudner @robwise

robwise:
Cool! It was unclear from the React documentation, but if you use Immutable objects (which we are), do you still need to do the `forceUpdate()` thing because of the shallow object introspection problem? I couldn't tell if they meant you didn't need it with Immutable objects or just that Immutable objects would make `forceUpdate()` run faster.

justin808:
@robwise I don't think forceUpdate() is necessary. It should just work.

justin808:
@mamute want to give this a try?

mamute:
sure :+1:

mamute:
Hey guys, how can I check if a component is rendering faster or any other result of pure-render implementation?

justin808:
@alexfedoseev @josiasds @alexfedoseev Should we be including the pure render mixin like this: https://christianalfoni.github.io/react-webpack-cookbook/Javascript-next.html

robwise:
@justin808 @mamute @alexfedoseev 

For performance checking, see https://facebook.github.io/react/docs/perf.html

PR #174, opened 2015-12-09
Include pure-render-decorator

mamute:
Including pure render decorator as proposed in #141 and organizing imports.


client/app/bundles/Comments/components/CommentForm.jsx
```diff
@@ -73,7 +76,7 @@ class CommentForm extends React.Component {
     if (this.state.formMode < 2) {
       ref = this.refs.text.getInputDOMNode();
     } else {
-      ref = React.findDOMNode(this.refs.inlineText);
+      ref = ReactDOM.findDOMNode(this.refs.inlineText);
```
mamute: This fixes a warning from React 0.14

client/app/bundles/Comments/components/SimpleCommentScreen.jsx
```diff
@@ -1,17 +1,20 @@
...
-import _ from 'lodash';
 
+@pureRender
 class SimpleCommentScreen extends React.Component {
   constructor(props, context) {
     super(props, context);
     this.state = {
       $$comments: Immutable.fromJS([]),
-      ajaxSending: false,
+      isSaving: false,
```
mamute: Using `isSaving` instead of `ajaxSending` to fix a warning at Inline Form on Simple React page, reflecting changes on #158

justin808: `fix a warning at Inline Form`

What was the warning?

justin808:
@mamute This looks quite good! 

@alexfedoseev Please take a peek at this! Will this break with Babel 6? @mamute would you mind trying Babel 6. Supposedly, they just fixed some things that are critical to make this work. @alexfedoseev can tell you more.

client/app/bundles/Comments/components/SimpleCommentScreen.jsx
mamute: Warning: Failed propType: Required prop `isSaving` was not specified in `CommentForm`. Check the render method of `SimpleCommentScreen`. And this kept the button enabled while saving...

justin808: @mamute Excellent. Yes, SimpleCommentScreen depends on CommentForm which was updated.

justin808: :+1:  Too bad we didn't see this in our tests!

alexfedoseev:
Yesterday we discussed it and decided to stick with [this solution](https://christianalfoni.github.io/react-webpack-cookbook/Javascript-next.html) for now, until decorators specs will be on stage 0 at least. @mamute will update this PR.

mamute:
@justin808 @alexfedoseev this is ready

alexfedoseev:
@mamute The idea is to create a parent `Component` with `shouldComponentUpdate` method:

```js
/* Component.jsx */

class Component extends React.Component {
  shouldComponentUpdate() {
    return React.addons.PureRenderMixin.shouldComponentUpdate.apply(this, arguments);
  }
}
```

And then:

```js
import Component from './Component';

export default class EachComponentInApp extends Component {

  // using `shouldComponentUpdate` from Component here...

}
```

mamute:
@alexfedoseev Done :+1:

alexfedoseev:
@mamute Looks good :+1:  One thing: let's place it in `client/app/libs/components` so it will be shared among all bundles.

mamute:
@alexfedoseev done :+1:

justin808:
@mamute This is awesome. One last question. Is there any way we can have a test or something that confirms we're getting some benefit from this change?

PR #202, opened 2016-01-18
Remove incorrectly-added dependency from top-level package.json

thewoolleyman:


justin808:
Thanks @thewoolleyman

PR #205, opened 2016-01-21
Adds Capybara.asset_host

mapreal19:
- Loads CSS from running server. Useful when debugging feature tests
with `save_and_open_page`.

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -6,6 +6,12 @@ require "rspec/rails"
 require "capybara/rspec"
 require "capybara/poltergeist"
 require "capybara-screenshot/rspec"
+
+# This will insert a <base> tag with the asset host into the pages created by
+# save_and_open_page, meaning that relative links will be loaded from the
+# development server if it is running.
+Capybara.asset_host = "http://localhost:4000"
+
 # Add additional requires below this line. Rails is not loaded until this point!
 
 # Requires supporting ruby files with custom matchers and macros, etc, in


mapreal19:
Before: 
image

After:
image

Easier to debug :dancer:

justin808:
:+1: :tada: :+1: :tada: 

SWEET!

CC: @geoffevason, @robwise @samnang

robwise:
@mapreal19 nice!!

spec/rails_helper.rb
```diff
@@ -6,6 +6,12 @@
 require "capybara/rspec"
 require "capybara/poltergeist"
 require "capybara-screenshot/rspec"
+
+# This will insert a <base> tag with the asset host into the pages created by
+# save_and_open_page, meaning that relative links will be loaded from the
+# development server if it is running.
+Capybara.asset_host = "http://localhost:4000"
```
justin808: @mapreal19 Why is this port 4000 and not port 3000? 

We switched to using port 3000 for the rails server, right? and port 4000 is for the hot reload server? And how does it work given that assets during development are coming from port 3500?

geoffevason:
Awesome!

spec/rails_helper.rb
mapreal19: yes you're right! I had an old version of the code... Fixed on #210 

It works including a base tag: `<base href="http://localhost:3000">`. That way `<link rel="stylesheet" media="all" href="/assets/application_static.css" data-turbolinks-track="true">` returns `localhost:3000/assets/application_static.css`

PR #216, opened 2016-02-03
Fix Procfiles to use new process names

robwise:
Basically we just don't need the special `build:test:client` script anymore, and you can just use `Procfile.static` to keep your static assets up to date while running your tests.

Also changed the ordering of the scripts so that the client and server versions of the same type of script are next to each other.

[Review on Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/216)

### DIFF
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -7,14 +7,13 @@
 rails: REACT_ON_RAILS_ENV=HOT rails s
 
 # Run the hot reload server for client development
-rails-assets: sh -c 'rm app/assets/webpack/* || true && HOT_RAILS_PORT=3500 npm run build:dev:client'
+hot-assets: sh -c 'rm app/assets/webpack/* || true && HOT_RAILS_PORT=3500 npm run hot-assets'
 
-# Keep the JS fresh for server rendering. Remove if not server rendering
-rails-server-assets: sh -c 'npm run build:dev:server'
+# Render static client assets
+rails-static-client-assets: sh -c 'npm run build:dev:client'
 
-# If you don't keep this process going, you will rebuild the assets per spec run. This is configured
-# in rails_helper.rb.
-rails-test: sh -c 'npm run build:test:client'
+# Render static client assets. Remove if not server rendering
+rails-static-server-assets: sh -c 'npm run build:dev:server'
 
 # Run an express server if you want to mock out your endpoints. No Rails involved!
 # Disable this if you are not using it.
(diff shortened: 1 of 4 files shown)


Procfile.dev
```diff
@@ -12,10 +12,6 @@ rails-assets: sh -c 'rm app/assets/webpack/* || true && HOT_RAILS_PORT=3500 npm
 # Keep the JS fresh for server rendering. Remove if not server rendering
 rails-server-assets: sh -c 'npm run build:dev:server'
 
-# If you don't keep this process going, you will rebuild the assets per spec run. This is configured
-# in rails_helper.rb.
-rails-test: sh -c 'npm run build:test:client'
```
justin808: we need this

please don't remove

this is used b/c above we do hot ones

Procfile.spec
```diff
@@ -1,4 +0,0 @@
-# For keeping webpack bundles up-to-date during a testing workflow.
-# If you don't keep this process going, you will rebuild the assets per spec run. This is configured
-# in rails_helper.rb.
-rails-test: sh -c 'npm run build:test:client'
```
justin808: we need this

Procfile.static
```diff
@@ -1,4 +1,5 @@
-# Run Rails without hot reloading (static assets)
+# Run Rails without hot reloading (static assets). Useful for running during
+# tests.
 rails: rails s
```
justin808: not tru

package.json
```diff
@@ -15,10 +15,10 @@
     "build:clean": "rm app/assets/webpack/*",
     "build:production:client": "(cd client && npm run build:production:client --silent)",
     "build:production:server": "(cd client && npm run build:production:server --silent)",
-    "build:dev:client": "(cd client && npm run build:dev:client --silent)",
     "build:client": "(cd client && npm run build:client --silent",
-    "build:dev:server": "(cd client && npm run build:dev:server --silent)",
     "build:server": "(cd client && npm run build:server --silent)",
+    "build:dev:client": "(cd client && npm run build:dev:client --silent)",
+    "build:dev:server": "(cd client && npm run build:dev:server --silent)",
     "hot-assets": "(cd client && npm run hot-assets)",
```
justin808: did you only change the order of the lines?

robwise: @justin808 correct

Procfile.dev
justin808: this is still sort of necessary

PR #224, opened 2016-02-09
Upgrade version of react-on-rails to 3

jbhatab:
[Review on Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/224)


Gemfile
```diff
@@ -37,7 +37,7 @@ gem "sdoc", group: :doc
 # Use Rails Html Sanitizer for HTML sanitization
 gem "rails-html-sanitizer"
 
-gem "react_on_rails"
+gem "react_on_rails", git: "https://github.com/shakacode/react_on_rails.git", tag: "3.0.0-beta.1"
```
justin808: why do this when the gem is on rubygems?

justin808:
Need to get CI to pass, and use rubygems version of beta.

---

Reviewed 4 of 4 files at r1, 13 of 13 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react-webpack-rails-tutorial/224#-:-KA4BQoCJtvq5vj3BPh5:871269859)*

justin808:
@jbhatab Any updates on this one?

Gemfile
jbhatab: @justin808 You need git defined to use tag.

justin808:
@alexfedoseev If you can, take a peek at this one. Odd that we get the error here with react-router.

justin808:
@jbhatab I investigated with Alex. 

We should update to react-router v2 final and double check our API usage. The problem is probably due to the beta version that in the tutorial.



to update the deps, 

```
rm npm-shrinkwrap.json
ncu -a -u
npm i
```

client/app/bundles/comments/store/routerCommentsStore.js
```diff
@@ -0,0 +1,29 @@
...
+  const initialState = {
+    $$commentsStore: $$commentsState.merge({
+      $$comments: initialComments,
+    }),
+  };
+
+  // https://github.com/reactjs/react-router-redux
+  const reducer = combineReducers({
+    ...reducers,
+    routing: routerReducer,
```
justin808: @alexfedoseev does it make sense to have a different store for the client with router, as compared to the server store? Maybe it doesn't matter?

Possibly the routeReducer would expect the history.

client/app/libs/middlewares/loggerMiddleware.js
```diff
@@ -8,11 +8,13 @@ export default function logger({ getState }) {
...
-    const immutableState = getState();
+    const state = getState();
     const readableState = {};
-    for (const storeItem in immutableState) {
-      if (immutableState.hasOwnProperty(storeItem)) {
-        readableState[storeItem] = immutableState[storeItem].toJS();
+    for (const storeItem in state) {
+      if (state.hasOwnProperty(storeItem)) {
+        readableState[storeItem] = (
+          state[storeItem].toJS ? state[storeItem].toJS() : state[storeItem]
```
justin808: @alexfedoseev This is b/c the react-router-redux state is a plain object.

client/app/bundles/comments/store/routerCommentsStore.js
alexfedoseev: I don't think we need different store. This is `routeReducer`:

```js
// Reducer

const initialState = {
  location: undefined
}

export function routeReducer(state = initialState, { type, payload: location }) {
  if (type !== UPDATE_LOCATION) {
    return state
  }

  return { ...state, location }
}
```

client/app/libs/middlewares/loggerMiddleware.js
alexfedoseev: :+1:

client/app/bundles/comments/store/routerCommentsStore.js
justin808: I removed that file (local one) unless you're talking about the one from react-router-redux.

Probably doesn't hurt, right, to have both?

justin808: But maybe better to remove.

justin808: Since this PR is already merged, feel free to remove on the next PR, @alexfedoseev

PR #234, opened 2016-02-18
Minify js in production.

lukeasrodgers:
The rake asset compilation task will now run the production-specific npm build scripts, which employ the `UglifyJsPlugin` plugin.

There may be reasons I'm not aware of to *not* do this, but this is what I would assume would happen for a production build, which is generally what you're targeting when running asset precompilation.

The numbers below just approximately sum vendor and app bundles.

current `bin/rake assets:precompile` (total approx 6.1 MB)

precompile-current

explicitly running npm prod builds in `bin/rake assets:precompile` (total approx 2.3 MB)

precompile-current-prod

explicitly running npm prod builds in `bin/rake assets:precompile` AND adding minification plugin (total approx 600 KB)

precompile-witih-min

[Review on Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/234)

### DIFF
--- a/lib/tasks/assets.rake
+++ b/lib/tasks/assets.rake
@@ -13,8 +13,8 @@ namespace :assets do
 
   desc "Compile assets with webpack"
   task :webpack do
-    sh "cd client && npm run build:client"
-    sh "cd client && npm run build:server"
+    sh "cd client && npm run build:production:client"
+    sh "cd client && npm run build:production:server"
     sh "mkdir -p public/assets"
 
     # Critical to manually copy non js/css assets to public/assets as we don't want to fingerprint them


justin808:
2.3 MB ==> 600K

:clap: :clap: :clap:

lib/tasks/assets.rake
```diff
@@ -13,8 +13,8 @@ namespace :assets do
 
   desc "Compile assets with webpack"
   task :webpack do
-    sh "cd client && npm run build:client"
-    sh "cd client && npm run build:server"
+    sh "cd client && npm run build:production:client"
+    sh "cd client && npm run build:production:server"
```
justin808: **GREAT** catch!

client/webpack.client.rails.build.config.js
```diff
@@ -69,6 +69,9 @@ if (devBuild) {
   console.log('Webpack dev build for Rails'); // eslint-disable-line no-console
   config.devtool = 'eval-source-map';
 } else {
+  config.plugins.push(
+    new webpack.optimize.UglifyJsPlugin()
+  );
```
justin808: @lukeasrodgers Can you check if this is really needed. Since we're loading the client file through the asset pipeline, it should be minified.

**HOWEVER** the server side file might benefit from this plugin, since it does not go through the asset pipeline.

@alexfedoseev, thoughts?

alexfedoseev: It's absolutely needed for client bundles w/o assets pipeline, but I remember I've seen somewhere that re-parsing of minified builds might increase build time (I might mix up something), so I'm not sure about it w/ pipeline. For server side it can slightly improve performance, never measured it though.

lukeasrodgers: @justin808 you're right, that was was unneeded -- I was misunderstanding the fact that the rails js helpers will serve different bundles depending on prod/dev environment. 

The good news is that running in production with just that single change to the rake task results in even greater savings:

before (www.reactrails.com)
www

after
www-after

It would be good to have someone else verify that the reduction in file size is real -- lots of switching back and forth between branches, manually running commands, and taking screenshots is error-prone.

I'll remove the line for the webpack uglify plugin and squash those commits, if that sounds cool.

lukeasrodgers:
I removed the extraneous webpack uglify plugin code, so we're just relying on the rails asset pipeline for now. It sounds like further optimization may be possible, but I think that can be accomplished in other PRs.

client/webpack.client.rails.build.config.js
justin808: Super. I look forward to this!

PR #237, opened 2016-02-26
Automatically add all serverRegistration and clientRegistration files to webpack .

jDeppen:
One less step for the developer and [encourages modularity of domains (as found in the README)](https://github.com/shakacode/react_on_rails#understanding-the-organization-of-the-generated-client-code).

[Review on Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/237)

### DIFF
--- a/client/webpack.client.base.config.js
+++ b/client/webpack.client.base.config.js
@@ -21,7 +21,7 @@ module.exports = {
     ],
 
     // This will contain the app entry points defined by webpack.hot.config and webpack.rails.config
-    app: glob.sync("./app/bundles/**/startup/clientRegistration.js*"),
+    app: glob.sync('./app/bundles/**/startup/clientRegistration.js*'),
   },
   resolve: {
     extensions: ['', '.js', '.jsx'],
--- a/client/webpack.client.base.config.js
+++ b/client/webpack.client.base.config.js
@@ -3,6 +3,7 @@
 const webpack = require('webpack');
 const path = require('path');
 const autoprefixer = require('autoprefixer');
+const glob = require('glob');
 
 const devBuild = process.env.NODE_ENV !== 'production';
 const nodeEnv = devBuild ? 'development' : 'production';
@@ -20,9 +21,7 @@ module.exports = {
     ],
 
     // This will contain the app entry points defined by webpack.hot.config and webpack.rails.config
-    app: [
-      './app/bundles/comments/startup/clientRegistration',
-    ],
+    app: glob.sync("./app/bundles/**/startup/clientRegistration.js*"),
   },
   resolve: {
     extensions: ['', '.js', '.jsx'],
--- a/client/webpack.server.rails.build.config.js
+++ b/client/webpack.server.rails.build.config.js
@@ -13,7 +13,7 @@ module.exports = {
   context: __dirname,
   entry: [
     'babel-polyfill',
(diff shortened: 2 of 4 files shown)


justin808:
@jDeppen, Thanks! Interesting approach. I see the "Railsy Convention over Configuration" approach here. 

I'm probably OK with this. I'd like to get some other reviewers to comment as well.

This can definitely make the generators a bit easier to build, @jbhatab @aaronvb as we don't have to add another directory every time.

Thanks again!

@alexfedoseev @josiasds Thoughts?

---

Reviewed 4 of 4 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react-webpack-rails-tutorial/237#-:-KBV5gnnVzJvkba8V8om:-218931066)*

jDeppen:
Another option is to just look for `clientRegistration.js*` and `serverRegistration.js*` files rather than hard-coding the rest.

Seems like it would be nice to add/remove folders in "bundles" and limit outside configuration as much as possible.

Thanks.

---

Review status: 2 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react-webpack-rails-tutorial/237#-:-KBVUN1sVfe-2QOl9V_l:1213322066)*

justin808:
@jDeppen @jbhatab We might be changing to a pattern of just using one `registration.jsx` file if server and client are the same. We might change the glob to match that. Stay tuned.

client/webpack.client.base.config.js
```diff
@@ -20,9 +21,7 @@ module.exports = {
     ],
 
     // This will contain the app entry points defined by webpack.hot.config and webpack.rails.config
-    app: [
-      './app/bundles/comments/startup/clientRegistration',
-    ],
+    app: glob.sync('./app/bundles/**/startup/clientRegistration.js*'),
```
justin808: @alexfedoseev @robwise Too much magic, or something we'd use?

client/webpack.server.rails.build.config.js
```diff
@@ -12,7 +13,7 @@ module.exports = {
   context: __dirname,
   entry: [
     'babel-polyfill',
-    './app/bundles/comments/startup/serverRegistration',
+    ...glob.sync('./app/bundles/**/startup/serverRegistration.js*'),
```
justin808: @alexfedoseev @robwise Too much magic, or something we'd use?

client/webpack.client.base.config.js
alexfedoseev: It's ok if we want to compile all bundles into single output, but usually it looks like this:

```js
someBundle: './app/bundles/someBundle/startup/clientRegistration.js',
someOtherBundle: './app/bundles/someOtherBundle/startup/clientRegistration.js',
```

So globbing doesn't help much.

justin808:
@jDeppen I spoke with Alex about this.

This is probably not a best practice because many apps will want to split up the bundles. The advantage of having separate bundles is that you can change one bundle, and the others can stay cached in the browser.

that is why we separate the vendor-bundle and the app-bundle

That being said, I think for a smaller app, you won't have many bundles, and it's just as well to manually list multiple registration files if you have that.

PR #246, opened 2016-03-06
Fix comments json format

justin808:
Updated props to have comments top level prop of comments.

Now, the server responds:

    { comments: [ c1, c2, etc.] }

Before the props were just:

    [c1, c2, etc.]


justin808:
@robwise or @alexfedoseev Give this the LGTM and hit the merge button please!

PR #260, opened 2016-03-28
Update to 5.0.0.rc.1 of ReactOnRails

justin808:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/260)

### DIFF
--- a/client/app/bundles/comments/startup/ServerRouterApp.jsx
+++ b/client/app/bundles/comments/startup/ServerRouterApp.jsx
@@ -5,12 +5,13 @@ import { match, RouterContext } from 'react-router';
 import createStore from '../store/commentsStore';
 import routes from '../routes/routes';
 
-export default (props, location) => {
+export default (props, railsContext) => {
   const store = createStore(props);
 
   let error;
   let redirectLocation;
   let routeProps;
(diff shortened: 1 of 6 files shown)


client/app/bundles/comments/startup/ServerRouterApp.jsx
```diff
@@ -5,12 +5,13 @@ import { match, RouterContext } from 'react-router';
...
 import routes from '../routes/routes';
 
-export default (props, location) => {
+export default (props, railsContext) => {
   const store = createStore(props);
 
   let error;
   let redirectLocation;
   let routeProps;
+  const location = { railsContext };
```
justin808: The reason for the major version bump.

client/app/bundles/comments/startup/ServerRouterApp.jsx
```diff
@@ -5,12 +5,13 @@ import { match, RouterContext } from 'react-router';
 import createStore from '../store/commentsStore';
 import routes from '../routes/routes';
 
-export default (props, location) => {
+export default (props, railsContext) => {
```
justin808: The reason for the major version bump.

PR #261, opened 2016-03-28
Show Redux Dev Tools in a New Window

seoyoochan:
Show Redux Dev Tools in a New Window.

> Personal preferences vary, and whether to put the DevTools in a separate window, in a dock, or right inside you app’s user interface, is up to you. Make sure to check the documentation for the monitors you use and learn about the different props they support for customizing the appearance and the behavior.
> 
> Note that there are no useful props you can pass to the DevTools component other than the store. The store prop is needed if you don’t wrap <DevTools> in a <Provider>—just like with any connected component. To adjust the monitors, you need to pass props to them inside DevTools.js itself inside the createDevTools() call when they are used.

Learn More [**WalkThrough about redux-devtools**](https://github.com/gaearon/redux-devtools/blob/master/docs/Walkthrough.md)

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/261)


justin808:
I think this wasn't passing CI -- I'm going to revert this and we'll redo. We should have the PR for this serve as a clear example of adding this feature and avoid any extraneous changes.

justin808:
@seoyoochan Let's resubmit this, and make sure we can refer to this in the docs. Single commit and ONLY related to redux dev tools. I'm the only one that will merge to master.

PR #262, opened 2016-03-28
Update README with instructions for installing qmake

audionerd:
On a fresh install on OSX, `capybara-webkit` will fail to build.

```sh
Installing capybara-webkit 1.8.0 with native extensions
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.
[..]
Command 'qmake' not available
```

`capybara-webkit` is required only by the Gemfile's test group, so one option is to simply exclude that test group when installing.

I think a better option is to provide instructions for how to get `qmake` installed. The actual commands required on OSX are:

```sh
brew tap homebrew/versions
brew install qt55
brew link --force qt55
```

But since those might change, and vary for other operating systems, I recommend linking to the `capybara-webkit` install docs directly in the README.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/262)

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ See package.json and Gemfile for versions
 1. Check that you have Ruby 2.2.4 or greater
 1. Check that you're using the right version of node. Run `nvm list` to check. Use 5.5 or greater.
 1. Check that you have Postgres installed. Run `which postgres` to check. Use 9.4 or greater.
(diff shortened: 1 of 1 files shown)


justin808:
:lgtm:

---

Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/shakacode/react-webpack-rails-tutorial/262#-:-KE-qPnvXxF5Op57Dptu:1417088721)*

PR #284, opened 2016-04-30
update heroku documentation

roxolan:
Looks like currently there are three buildpacks being run from the .buildpacks file rather than from a directory.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/284)

### DIFF
--- a/Procfile.hot
+++ b/Procfile.hot
@@ -1,6 +1,6 @@
 # Basic procfile for dev work.
 # Runs all processes. Development is faster if you pick one of the other Procfiles if you don't need
-# some of the processes: Procfile.rails or Procfile.express
+# some of the processes: Procfile.hot or Procfile.express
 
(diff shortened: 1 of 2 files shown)


justin808:
There should be two: ruby and node.

Why do you think three?

and we're not using the .buildpacks file any more.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react-webpack-rails-tutorial/284#-:-KGkf5GnKLW0UA6Q8Trm:1670076427)*

roxolan:
I was thinking three because, in addition to ruby & node there was also a link to heroku-buildpack-sourceversion. But since it's not used anymore, there is no big deal, I am closing this.

PR #298, opened 2016-07-04
Fix server-express startup

milushov:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/298)


milushov:
Because we are using new es6 function syntax like `(req, res) => {}`

justin808:
@milushov Thanks! I'll try to test this out today.

justin808:
@milushov I just updated all dependencies. Please rebase your changes on top of that and confirm your fix works.

justin808:
See #298

PR #302, opened 2016-07-11
Enable serving static assets in development mode

techpeace:
I noticed that I wasn't able to use `image_tag` within a view. Just needed to activate static asset serving in development mode.

---

This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/302)

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -11,6 +11,9 @@ Rails.application.configure do
 
   # Show full error reports.
   config.consider_all_requests_local = true
+  
+  # Enable serving static assets
+  config.public_file_server.enable
 
   # Enable/disable caching. By default caching is disabled.
   if Rails.root.join("tmp/caching-dev.txt").exist?


config/environments/development.rb
```diff
@@ -11,6 +11,9 @@
 
   # Show full error reports.
   config.consider_all_requests_local = true
+  
+  # Enable serving static assets
+  config.public_file_server.enable
```
justin808: Per the the latest rails docs: [railsguides](http://edgeguides.rubyonrails.org/configuring.html)

config.public_file_server.enabled configures Rails to serve static files from the public directory. This option defaults to true, but in the production environment it is set to false because the server software (e.g. NGINX or Apache) used to run the application should serve static files instead. If you are running or testing your app in production mode using WEBrick (it is not recommended to use WEBrick in production) set the option to true. Otherwise, you won't be able to use page caching and request for files that exist under the public directory.

@techpeace Are you sure we need this?

techpeace:
You're absolutely right, I just assumed the docs were out of step with the final release. In a fresh Rails 5 app, this isn't necessary.

The sample app is suffering from this problem, though, so it must be something particular to the asset set-up. For some reason, calling `config.public_file_server.enable` enables the use of `image_tag`. Without it, all images appear as broken links. You can see what I'm referring to in [this example branch](https://github.com/techpeace/react-webpack-rails-tutorial/pull/1/files).

justin808:
@techpeace AWESOME! Please submit your whole branch as a PR, and move the image to the bottom next to the other styling images.

justin808:
@techpeace Any chance that you can finish this one up? If not, let me know.

justin808:
@techpeace, Any update?

config/environments/development.rb
justin808: @techpeace The docs say that the default is true. So it's not clear to me on why we need to set this for `development.rb`.

techpeace: I'm not sure why you need it, either, but if you'll check out my branch, you'll see that you do. Otherwise, static images don't render properly.

I actually won't have time to finish this up, as it looks like I won't be using this project in the near future.

justin808:
I'm going to close this one for now, as it's not clear on why this change is required.

techpeace:
Try to use `image_tag` within a view, and you'll get a broken image. That was why I submitted the PR.

justin808:
@techpeace:

1. Is the image path under `/client` or `/app/assets`?
2. How are you handling fingerprinting of the image?

I bet this is related to how webpack fingerprints assets plus with how your development asset path is setup.

If you can just make a change to the PR that shows that this is needed, I'll merge this. That means adding some small image to the project.

Here's one:

fng-logo

thanks!

techpeace:
Cool, thanks! I'll check this out shortly.

justin808:
@techpeace Any update?

PR #310, opened 2016-07-29
[Possible No-op] Adjust Procfiles for compatibility with both foreman and forego

zph:
First off, thanks for all your work on this project :).

forego doesn't allow - in the name of processes. This allows both
foreman and forego to work with these procfiles.

Link to regex in forego looks like it should work for dashes:
https://github.com/ddollar/forego/blob/master/procfile.go#L12. But
testing on my OSX dev machine using forego 0.16.1 (latest) shows it
fail to parse any of them beyond the `rails` entry. I also tried
rearranging order of entries in Procfiles, with same issue.

I'll take a look in forego to see if it should be fixed upstream, but
wanted to log this here in case anyone else get stumped for a bit trying
to debug this :).

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/310)

### DIFF
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -7,13 +7,13 @@
 rails: REACT_ON_RAILS_ENV=HOT rails s -b 0.0.0.0
 
 # Run the hot reload server for client development
-hot-assets: sh -c 'rm app/assets/webpack/* || true && HOT_RAILS_PORT=3500 npm run hot-assets'
+hot_assets: sh -c 'rm app/assets/webpack/* || true && HOT_RAILS_PORT=3500 npm run hot-assets'
 
 # Render static client assets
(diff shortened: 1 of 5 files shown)


justin808:
@zph this should probably be fixed on forego.

zph:
Filed in forego: https://github.com/ddollar/forego/issues/87

With forego's master copy, the existing Procfiles work if the rails entry is also wrapped in a `sh -c`. With master's forego, the dashes work ^_^.

Closing and will re-open if there's something actionable here.

Mahalo.

zph:
forego, by same author as foreman, broke compatibility.

It wasn't working correctly in 0.16.1, it does work correctly in master and I've asked them to cut a release from master.

PR #326, opened 2016-09-01
Update dev server command to use rails port variable

wtachau:
As the code is now, it ignores the local variable that can be set from any environment variables

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/326)


justin808:
Thanks! @wtachau You must be the first one to try changing this! 😄

Issue #331, opened 2016-10-24
Demonstrate this in a docker container


EDeijl:
It would be great if this project has a Dockerfile
It is a pretty complete application that a lot of people could learn from

sub2u:
I will push the Dockerfile for this project soon.

sub2u:
Created the pull request https://github.com/shakacode/react-webpack-rails-tutorial/pull/332

justin808:
This one almost made it to the finish line. It does need updating. Any takers?

thetonyrom:
I'll handle it.

thetonyrom:
Hi @justin808 . I've just submitted an initial dockerized version, have a look.

Trip4077:
If this still needs work, I will be happy to look at it as a part of my application process.

justin808:
See #507

PR #332, opened 2016-10-24
Docker support to run the project

sub2u:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/332)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM sub2u/ruby-2.3.1-node-6.8.0
+MAINTAINER sub2u
+
+COPY ./ app/
+
+WORKDIR /app
+
+RUN bundle install
+
+RUN npm install
+
+EXPOSE 5000
+
+CMD foreman start -f Procfile.hot
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM sub2u/ruby-2.3.1-node-6.8.0
 MAINTAINER sub2u
 
-COPY ./ app/
+ADD ./ app/
 
 WORKDIR /app
 
--- a/README.md
+++ b/README.md
@@ -84,6 +84,13 @@ See package.json and Gemfile for versions
 1. [Heroku for Rails 5 deployment](https://devcenter.heroku.com/articles/getting-started-with-rails5)
 1. [Turbolinks 5](https://github.com/turbolinks/turbolinks)
 
+## Basic Demo Setup with docker
+1. `docker pull sub2u/react-webpack-rails-tutorial`
+2. `run -d -p 5000:5000 sub2u/react-webpack-rails-tutorial`
+3.
+  1. Open a browser tab to http://localhost:5000 for the Rails app example.
+  2. When you make changes, you have to refresh the browser page.
+  
 ## Basic Demo Setup
 1. Be sure that you have Node installed! We suggest [nvm](https://github.com/creationix/nvm), with node version `v6.0` or above. See this article [Updating and using nvm](http://forum.shakacode.com/t/updating-and-using-nvm/293).
 1. `git clone git@github.com:shakacode/react-webpack-rails-tutorial.git`


sub2u:
Referring to https://github.com/shakacode/react-webpack-rails-tutorial/issues/331

justin808:
@dylangrafmyre Can you please review this one.

dylangrafmyre:
@sub2u Could you please link to your `sub2u/ruby-2.3.1-node-6.8.0` Dockerfile. I do not see it on Docker Hub. Is it in a Github repo?

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react-webpack-rails-tutorial/332#-:-KUsWRNlfYMSx1mOi3Sb:b-tp2mgh)*

sub2u:
@dylangrafmyre You can find the docker in docker hub
https://hub.docker.com/r/sub2u/ruby-2.3.1-node-6.8.0/

dylangrafmyre:
@sub2u there is no `Dockerfile` tab under your referenced docker hub repo.


---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react-webpack-rails-tutorial/332#-:-KUttW28o-QyLVFpsK4l:bqzbj8q)*

sub2u:
@dylangrafmyre Docker file is in github repo please review in https://github.com/sub2u/ruby-2.3.1-node-6.8.0

Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM sub2u/ruby-2.3.1-node-6.8.0
+MAINTAINER sub2u
+
+COPY ./ app/
```
alexfedoseev: Wouldn't it be more convenient to mount directory instead of coping over all the things inside container?

alexfedoseev:
@sub2u IMO you have too many layers in https://github.com/sub2u/ruby-2.3.1-node-6.8.0

Your stuff will be outdated if you will change something below `RUN apt-get update -qq ...` as this layer is cached by the docker and will be reused.

Dockerfile
sub2u: Sure, we can use ADD, i will update it

sub2u:
@alexfedoseev we can use force update, in `RUN apt-get update -qq` but j am not sure if will effect the build performance it always try to get updates instead of fetching from caching layer.

Dockerfile
alexfedoseev: I was meaning not `ADD`, but [volumes](https://docs.docker.com/engine/tutorials/dockervolumes/#/mount-a-host-directory-as-a-data-volume). Copying process takes time and in development copy files on change is kinda pricy. If volume is used, then it reads the files directly from the mounted volume, so it should be faster.

alexfedoseev:
> we can use force update, in RUN apt-get update -qq but j am not sure if will effect the build performance it always try to get updates instead of fetching from caching layer.

If it acts like this then I believe there's something wrong is going on w/ this setup as unchanged layer must be taken from the cache: https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#/minimize-the-number-of-layers

dylangrafmyre:
@sub2u Thank you for taking the time to work on this. For this project I think it would be best to have a project Dockerfile  that we maintained off of the the Ruby 2.3.1-alpine base image or similar. I  do not think it is best to rely on a maintained image by contributors to this project. 

@alexfedoseev brings up very good points about image size and minimizing image layers. 

I would be more than happy to work on getting the right Dockerfile config and build in place with you that meets the needs of this tutorial.

---

Reviewed 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react-webpack-rails-tutorial/332#-:-KUuJbtl_TcgA6H3x0x4:b-fisakc)*

sub2u:
Sure @dylangrafmyre i will update the docker file

sub2u:
Yes agreed Image size should be less,


It contain All Ruby nodejs, and other libraries required for project it is 366 MB size.
I will prepare the docker file with volume to link from external volumes.

justin808:
@sub2u Any update?

justin808:
@sub2u Anything I can do to help?

PR #337, opened 2016-11-05
Code splitting with server rendering example

jtibbertsma:
Example of using async routes with server rendering.

Rather than altering the current react-router example, I added new code with a separate Procfile, and separate webpack config with a different entry point. My reasoning for doing this is that people looking for a react-router example might not want to try using code splitting just yet.

So, I mostly added new code without altering any existing code. The existing code that I changed was moving the function `checkAuth` into its own file. The problem with having `checkAuth` defined as a static method on `TestReactRouterRedirect` is that `TestReactRouterRedirect` gets fetched asynchronously, but `checkAuth` needs to be defined in app-bundle so that it can be passed to the onEnter hook in the routes.

See https://github.com/shakacode/react_on_rails/pull/581

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/337)


justin808:
@jtibbertsma Should this PR be merged? If so, can you please rebase on top of master?

justin808:
@jtibbertsma Should we be merging this one?

@tmobaird, FYI

jtibbertsma:
No, this probably shouldn't be merged. I should probably rewrite it in
webpack v2 and react-router v4 before it gets merged, and I probably won't
have time for that anytime soon. You can close it if you want

justin808:
@jtibbertsma I'll mark this one as up for grabs. @tmobaird might be interested.

justin808:
Code splitting and dynamic imports with Webpack v4 with React on Rails are formally supported as part of [React on Rails Pro](https://github.com/shakacode/react_on_rails/wiki) and you can see this live, in-action on https://egghead.io. [Email me for more info](mailto:justin@shakacode.com).

PR #349, opened 2016-12-09
React native client

justin808:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/349)


justin808:
@alleycat-at-git I  will update this PR per our discussions with @robwise and @alexfedoseev. As a useful example to contrast thunks, we might use redux-saga. @robwise, any opinion on that one?

robwise:
@justin808 Although I like Saga, I thought we had decided to just use plain thunks so it would be easier for people to understand

PR #350, opened 2016-12-17
Remove mini_racer as a dependency

kennethkalmer:
Since we depend on nodejs properly, we can just let execjs use node instead of pulling in the huge libv8 dependency

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/350)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -39,10 +39,6 @@ gem "rails-html-sanitizer"
 
 gem "react_on_rails", "~> 6.1"
 
-# See https://github.com/sstephenson/execjs#readme for more supported runtimes
-# mini_racer is probably faster than therubyracer
-gem "mini_racer"
-
 gem "autoprefixer-rails"
 
 gem "awesome_print"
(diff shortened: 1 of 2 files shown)


kennethkalmer:
This PR came a bit prematurely, I wanted to trigger a travis-ci run against my fork... But seeing it passed we can ship away at it (that is if anything else is required).

justin808:
@kennethkalmer mini_racer is required if we're using the rails server to do server rendering.

I encourage you to experiment with the node server and let us know how that goes. A few of us are working on a better solution with a node server. Let me know if you want a slack invite to join us.

kennethkalmer:
No sweat, I wanted to use this as a base for a meetup demo but the size of the dependencies was a deal breaker at the time... I was on crappy 3G away for the weekend and the bundler never finished the install due the large dependency, and when I removed it the demo just worked...

So that was the impetus for dropping the dependency, especially given the tests all passed without it...

You folks have done a lot of work to tame this beast, I'm sure a lot of people will benefit from the work. As for myself, I'm spoiled with Clojure & ClojureScript in my day-to-day, which makes this entire dance unnecessary. 

:beers:

Issue #382, opened 2017-04-05
Upgrade JS and Ruby Libraries


justin808:
I'd like to get all the JS and Ruby libraries updated.

Anybody want to help?

tmobaird:
@justin808 I can help out! Do you just want to update the gem/node module versions? or the ruby and node versions as well?

justin808:
Let's update all. Thanks @tmobaird!

tmobaird:
@justin808 did you want to jump to the most recent, stable versions for node (7.8.0) and ruby (2.4.1)?

justin808:
> @justin808 did you want to jump to the most recent, stable versions for node (7.8.0) and ruby (2.4.1)?

YES!

justin808:
Halfway done! #383!

tmobaird:
Currently waiting on the next release of react-bootstrap that is compatible with react v >= 15.5.0.

tmobaird:
Ok, It looks like react-bootstrap has finally updated everything for react v15.5. I will continue upgrading these dependencies tomorrow!

justin808:
@tmobaird FABULOUS!

tmobaird:
@justin808 sorry for the false alarm. It looks like react-bootstrap has a couple dependencies (react-overlays and uncontrollable) that have not been updated yet and are still using deprecated react syntax. Lets wait a couple more days on this

justin808:
@tmobaird want to try again? BTW: https://medium.com/webpack/webpack-3-official-release-15fd2dd8f07b 

Let's bump Webpack as well!

tmobaird:
@justin808 definitely! I'll knock this out asap! 👍

justin808:
@tmobaird Any progress?

tmobaird:
@justin808 actually yes! Sorry this has taken so long, but react router v4 is fundamentally different from the earlier versions so it required some extra work to update the router code for both client and server side.

I should have a PR up by today or tomorrow with updates for a lot of the main JS dependencies.

justin808:
@tmobaird, check out #443. We're also in the process of hiring right now, if you know anybody looking. Anybody interested can [email me](mailto:justin@shakacode.com).

PR #385, opened 2017-04-09
Maintenance/upgrade js dependencies

tmobaird:
References #382.

@justin808 here's the second PR for updating the dependencies. Note: **This PR does NOT include the update of the react libraries**. Apparently the new `react`, `react-router`, `react-dom` versions that were released added a lot of breaking changes. An example is removing `PropTypes` from the react library and adding it to it's own npm module. I'm not sure the amount of updates this will cause, but I figured it was best to leave out of this PR for size.

Any suggestions about how to tackle this?

This PR includes:
- Updating the node version in the `.nvmrc`
- Adding yarn to `engines` in `package.json`'s
- Updating JS dependencies

Also I noticed that npm was still included in the engines section of the `package.json`'s. Any idea if this can be removed?

Feel free to let me know any other suggestions you have!

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/385)


justin808:
https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html

@tmobaird I think we should try to upgrade. The codemod *should* make this painless.

justin808:
@tmobaird I figure it's best if we do the 15.5 one in its own PR.

tmobaird:
@justin808 awesome, that sounds good. I'll try to knock that out today! 👍

PR #386, opened 2017-04-11
Fix for Switch tab issue

richa:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/386)


justin808:
@richa What's the plan with this?

Issue #361, opened 2017-03-01
Possible to use React Hot Loader?


mattfordham:
Hi there!

Desipte the link in the [React on Rails Doctrine](https://blog.shakacode.com/the-react-on-rails-doctrine-3c59a778c724#.q8pcumtie) to [React Hot Loader](https://github.com/gaearon/react-hot-loader), it seems this project is using the deprecated [React Transform HMR](https://github.com/gaearon/react-transform-hmr) project.

Has anyone had success implementing React Hot Loader yet and/or have an example of implementation? Specifically, I am getting hung up on where/how to use the `module.hot.accept` stuff.

Thanks!

mattfordham:
FWIW, I think I figured it out a good way to do this...

```
import React from 'react';
import ReactOnRails from 'react-on-rails';
import { AppContainer } from 'react-hot-loader';

import HelloWorld from './HelloWorld';

const App = (props, _railsContext) => {
  if (module.hot) {
    module.hot.accept('./HelloWorld', () => {
      ReactOnRails.reactOnRailsPageLoaded();
    });
  }

  return (
    <AppContainer>
      <HelloWorld {...props} />
    </AppContainer>
  );
};

ReactOnRails.register({ App });
```

This seems to work, although I havn't tested thoroughly.

justin808:
@mattfordham Can you help me with a PR with these updates? I'd be glad to give you help!

mattfordham:
Yeah, I'd love to try! Won't me immediately, but will put on the list ;)

justin808:
This is related: https://github.com/shakacode/react_on_rails/pull/742

Judahmeek:
Also https://github.com/shakacode/react_on_rails/issues/332. I've been exploring this problem domain and I don't think `ReactOnRails.reactOnRailsPageLoaded()` is the best choice since it rerenders all registered components regardless of the scope of the hot update.

justin808:
@Judahmeek @mattfordham FWIW, the hot loader seems to work GREAT if you are not using CSS modules.

Judahmeek:
@justin, what about flexbox?

retroalgic:
@mattfordham 
[...]

PR #395, opened 2017-05-29
Upgrade react-webpack-rails-tutorial to Webpacker Lite

justin808:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/395)


kevinhughes27:
I had to add `HOT_RELOADING=TRUE` to my Procfile in addition to these changes to ensure webpacker_lite is also expecting reloading.

https://github.com/shakacode/webpacker_lite/blob/master/lib/webpacker_lite/env.rb#L14
https://github.com/shakacode/webpacker_lite/blob/master/lib/webpacker_lite/configuration.rb#L24

justin808:
@Judahmeek @kevinhughes27  is this something that should be added?

kevinhughes27:
I was initially expecting `REACT_ON_RAILS_ENV=HOT` to set this up for me. I think ideally we wouldn't set 2 ENV vars in the Procfile but that's just my opinion.

justin808:
@kevinhughes27, @judahmeek, @conturbo, we need to finish up the half completed work of moving to React on Rails v8:

1. Remove REACT_ON_RAILS_ENV, as that's no longer used for v8. See [search](https://github.com/shakacode/react-webpack-rails-tutorial/search?utf8=%E2%9C%93&q=REACT_ON_RAILS_ENV&type=)
2. Fix hot reloading in here: https://github.com/shakacode/react-webpack-rails-tutorial/issues/361

kevinhughes27:
I'm happy to PR the `REACT_ON_RAILS_ENV` changes to the example repo. I started implementing the changes in #361 but given the state of my project I need to continue using the old hot reloader since I can't have a root component until I finish more refactoring. I'm not sure how many others would be in that state or how much the focus of react_on_rails is for old apps vs new apps.

Issue #404, opened 2017-07-07
Basic Demo Setup instruction lists wrong local host location


dvcarter:
The Basic Demo Setup documentation states the following instructions:

...
14. Open a browser tab to **http://localhost:3000** for the Rails app example with HOT RELOADING
15. Try Hot Reloading steps below!
16. foreman start -f Procfile.static
17. Open a browser tab to **http://localhost:3000** for the Rails app example.
18. When you make changes, you have to refresh the browser page.

In both cases (number 14 and number 17), the wrong default localhost is listed. The default localhost is actually **http://localhost:5000**

justin808:
See my response: https://github.com/shakacode/react-webpack-rails-tutorial/pull/409#issuecomment-315683620

PR #409, opened 2017-07-16
fix localhost location

tjnet:
fix localhost location of README.

Please see #404 😃

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/409)

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,10 +148,10 @@ See package.json and Gemfile for versions
 1. `yarn`
 1. `rake db:setup`
 1. `foreman start -f Procfile.hot`
-  1. Open a browser tab to http://localhost:3000 for the Rails app example with HOT RELOADING
+  1. Open a browser tab to http://localhost:5000 for the Rails app example with HOT RELOADING
   2. Try Hot Reloading steps below!
 1. `foreman start -f Procfile.static`
-  1. Open a browser tab to http://localhost:3000 for the Rails app example.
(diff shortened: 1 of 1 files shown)


justin808:
@tjnet this line should be configuring port 3000:

https://github.com/shakacode/react-webpack-rails-tutorial/blob/8a6c8aa2e3b7ae5b08b0a9744fb3a63a2fe0f002/config/puma.rb#L12

Are you sure you don't have PORT defined in your environment to be 5000?

FWIW, I also define PORT in my environment to be 3000.

The default port for foreman to use with Rails is 5000.

tjnet:
> Are you sure you don't have PORT defined in your environment to be 5000?

No, I don't have.

>FWIW, I also define PORT in my environment to be 3000.

oh... I think that newbie can not notice this .
It might be better to write this 😈

justin808:
I'm not clear why you're getting port 5000 when we configure the port to be 3000 in the puma file.

---

Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/shakacode/react-webpack-rails-tutorial/409#-:-KpNC5QOOOY2kV8ThFZn:b-87h5xh)*

tjnet:
>FWIW, I also define PORT in my environment to be 3000.

I couldn't notice this is needed.

PR #418, opened 2017-09-21
README: Use HTTPS URL instead of SSH

rodrigobdz:
Using HTTPS is the [recommended way by Github](https://help.github.com/articles/which-remote-url-should-i-use/) when cloning repos.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/418)


justin808:
@rodrigobdz where does it say use HTTPS?

rodrigobdz:
In the [Cloning with HTTPS URLs](https://help.github.com/articles/which-remote-url-should-i-use/#cloning-with-https-urls-recommended) section.

justin808:
I prefer SSH.

PR #420, opened 2017-10-19
Updated RoR to version 10.0.0

Ifegwu:
WIP: Hot reload with the command `foreman start -f Procfile.hot`, browser takes long to load. When loaded assets disappear. For now we can only run the application using `foreman start -f Procfile.static` which works fine.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/420)


justin808:
See #421

PR #430, opened 2017-12-21
Updating Docs

RathanKumar:
Changed port where the server starts serving rails example application

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/430)

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,10 +148,10 @@ See package.json and Gemfile for versions
 1. `yarn`
 1. `rake db:setup`
 1. `foreman start -f Procfile.hot`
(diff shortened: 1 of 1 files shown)


justin808:
https://github.com/shakacode/react-webpack-rails-tutorial/blob/master/config/puma.rb#L12 sets the port to 3000 unless your ENV value for PORT is set.

Can you do a bit more research on this one?

RathanKumar:
Hi @justin808, Thanks for looking into this. I think I'm missing something. When I clone and start the project with `foreman start -f Procfile.hot` server is not serving on port `:3000` instead it's `:5000`. Can you please guide me with some hints to understand this.

PR #432, opened 2018-01-10
Fix Mocha debugging

reconstructions:
Running the `yarn test:debug` script in `client/package.json` fails with this error:

```sh
$ yarn test:debug
yarn test:debug v0.25.2
$ yarn run test -- --debug-brk
yarn run v0.25.2
$ NODE_PATH=./app mocha --require babel-core/register --require 
./app/libs/testHelper.js --require ./app/libs/testNullCompiler.js 'app/**/*.spec.@(js|jsx)'
 --debug-brk
(node:90448) [DEP0062] DeprecationWarning: `node --debug` and `node 
--debug-brk` are invalid. Please use `node --inspect` or `node --inspect-brk` instead.
error Command failed with exit code 9.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
```

Which is fixed by replacing `client/package.json` with `"test:debug": "yarn run test -- debug”`.  Using `debug` is also deprecated but it does work, the suggested `inspect` syntax fails with the current `test` script syntax.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/432)


justin808:
Let's rebase this on top of maaster.

justin808:
This isn't right either. We don't need the `--` for newer node/yarn and we should document the difference of --inspect, --inspect-brk

ramrudra:
@justin808 As there are plans to migrate from mocha to Jest, is this still valid?

justin808:
@ramrudra there's a PR I'm about to merge for jest.

ramrudra:
@justin808 In that case may be we should mark PRs like this with an appropriate label so devs can contribute to open issues than accidentally spending time on these. Just a suggestion.

PR #433, opened 2018-01-23
Fix webpack asset bundling for tests

reconstructions:
@justin808 I am using this to try and troubleshoot the CI problems.  They seem to be caused by asset compilation failures, and I have found a couple different places this is failing prior to test runs running locally.  

The package.json changes in this PR fixes one of them.  When running `(cd client && yarn run build:test)`, webpack runs and appears to succeed, but actually the bundles are not updated because the `rm` statement that is supposed to delete the contents of `public/webpack-test` had a syntax error and doesn't do its thing.

There are other places test run asset compilation isn't working, we'll see if this passes CI and I will take it from there.

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/433)

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -35,7 +35,7 @@
     "build:dev:client": "NODE_ENV=development webpack -w --config webpack.client.rails.build.config.js",
     "build:dev:server": "NODE_ENV=development webpack -w --config webpack.server.rails.build.config.js",
     "build:server": "webpack --config webpack.server.rails.build.config.js",
-    "build:test": "rm -rf ../public/webpack/test && NODE_ENV=test yarn run build:client && NODE_ENV=test yarn run build:server",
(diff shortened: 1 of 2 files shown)


justin808:
I'm not convinced these changes make sense.

reconstructions:
No worries, @justin808 whatever you think is OK, I can work on other issues.  

The `client/package.json` [change](https://github.com/reconstructions/react-webpack-rails-tutorial/blob/95afa07efa17d3bb1fa11e37cdf05d24d79eec0e/client/package.json#L38) fixes a typo in what seems to be a broken script (there is no directory named `public/webpack/test` to `rm`, the correct path is `public/webpack-test`.  Without this change the `yarn run build:test` runs but does not update the bundle, and as a result when changes are made, tests fail when they should succeed and vice-versa.  

Test assets are also not compiling as they should elsewhere, and I got some valuable info about what might be causing that from pushing the change to `package.json`, which otherwise doesn't make sense, agreed.  I think I will set up my own CI so I can dig a little deeper into this.

Issue #407, opened 2017-07-15
Remove application_static from docs due to Webpacker changes


justin808:
Per this search:

https://github.com/shakacode/react-webpack-rails-tutorial/search?utf8=%E2%9C%93&q=application_static&type=

This needs cleanup.

rnice01:
Hey I'd like to help out with this if I could, seems simple enough.

justin808:
@rnice01 Please start on this! I'll mark this taken by you, but you have only a few days to show progress. Get in touch if you want a Slack invite as well. And we need to update the docs for https://github.com/shakacode/react_on_rails.

ramrudra:
@justin808 I'll pick this up

singhvaibhavram:
@justin808 I would like to take this.

justin808:
@singhvaibhavram Sure. If we can get proper HMR working as well, that would be great.

ramrudra:
@justin808 @singhvaibhavram  PR #446 is for this specific issue.

singhvaibhavram:
Yup. Will try that too. Give me a day to get me acquainted with the project
and then will do that.!!

singhvaibhavram:
@ramrudra can you msg me on slack? I have to ask you something regarding the pull request you made. 
slack email: vaibhavnarayansingh147@gmail.com

singhvaibhavram:
@justin808 I have messaged you on slack for some doubts.

ramrudra:
@singhvaibhavram Sure. 

@justin808 Could you please send me a slack invite on ramrudra.p@gmail.com

justin808:
Closing as consolidated into the broader documentation cleanup in #525. The remaining `application_static` reference is still visible in `docs/jquery-with-react-on-rails.md` and should be handled as part of that current docs pass.

PR #446, opened 2018-08-26
#407 Remove application_static from docs

ramrudra:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/446)


justin808:
@ramrudra the fix is not just to delete this but replace with the correct instructions given the change to using rails/webpacker.

ramrudra:
@justin808 My bad. Just got confused with the issue title. We can ignore this PR for now as @singhvaibhavram would like to pick this up.

PR #454, opened 2019-01-28
Docker workaround

KenyStev:
This is a Work around for setting postgresql server and redis server up and running through docker containers

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/454)

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 ## React on Rails Pro and ShakaCode Pro Support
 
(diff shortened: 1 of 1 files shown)


justin808:
that’s good, but it’s not really a workaround
it’s just a good alternative to using brew or other installations so that you can control the versions of pg and redis for the app

PR #475, opened 2020-06-26
update react_on_rails to beta version

hussainakram:
Testing configuration changes: [justin808/remove-precompile-rake](https://github.com/shakacode/react_on_rails/tree/justin808/remove-precompile-rake)

---
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/475)


bin/webpack
```diff
@@ -1,3 +1,8 @@
 #!/bin/bash
-cd client && yarn run build:production
-cd client && yarn run build:test
+if [[ $RAILS_ENV = "production" ]] ; then
+  cd ../client && yarn run build:production
```
justin808, line 3: put the cd before the if statement

DRY

bin/webpack
```diff
@@ -1,3 +1,8 @@
...
+fi
+
+if [[ $RAILS_ENV = "test" ]] ; then
+  cd ../client && yarn run build:test
+fi
```
justin808, line 8: and we need the else for others, like development

and the ENV should be be NODE_ENV falling back to RAILS

look at the default config/webpack/production.js for webpacker installs

PR #507, opened 2022-10-31
Controlplane setup

dzirtusss:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/507)


.controlplane/controlplane.yml
```diff
@@ -0,0 +1,18 @@
+aliases:
+  common: &common
+    org: shakacode-staging
+    location: aws-us-east-2
+    one_off_workload: rails
+    app_workloads:
+      - rails
+    additional_workloads:
+      - redis
+      - postgres
```
justin808, line 10: @dzirtusss why do we nee to list these?

dzirtusss: those are used in `ps` command atm - `ps`, `ps:start`, `ps:stop`, `ps:restart`. however, there is no clear plan about their future need atm

Dockerfile
```diff
@@ -11,12 +11,12 @@ COPY Gemfile* ./
...
 RUN bundle install
 
-COPY package.json .
+COPY package.json yarn.lock .
 RUN yarn install
 
 COPY . .
 
-RUN rake react_on_rails:locale && RAILS_ENV=production rake assets:precompile
+RUN rails react_on_rails:locale && RAILS_ENV=production rails assets:precompile
```
justin808, line 19: Nice to set NODE_ENV=production as well.

Dockerfile
```diff
@@ -11,12 +11,12 @@ COPY Gemfile* ./
...
 
 ENV RAILS_ENV=production 
 ENV NODE_ENV=production
```
justin808, line 22: Move these lines above the precompile.

PR #529, opened 2023-06-19
chore: bump major versions of client dependencies

KhaledEmaraDev:
1. Bump all minor and patch versions of all packages.
2. Upgrade `autoprefixer`: Use `.browserslistrc` and delete `.postcssrc` add it to `postcss.config.js` . Install and add `postcss-preset-env` to `postcss.config.js`.
3. Upgrade `immutable`: Explicitly convert all `Array`s to `List`s.
4. Upgrae `marked`: Install `marked-gfm-heading-id` to support `gfm` option.
5. Upgrade `react-router`:
  a. Remove `exact`.
  b. Use `Routes` instead of `Switch`.
  c. Import `StaticRouter` from `react-router-dom/server`.
  d. Use `element` instead of `component`.
  e. Use absolute paths instead of relative in Links.
  f. Convert Two Components to Functional React ones in order to use `useLocation` hook.
  g. Use `isActive` in `className` instead of `activeClassName`.
  h. Use `Navigate` instead of `Redirect`.
  i. Reset state once redirected.
6. Upgrade `react-transition-group`:
  a. Use `TransitionGroup` and `CSSTransition` instead of `ReactCSSTransitionGroup`.
  b. Convert one component to a Functional React one to use `forwardRef`.
  c. Add `nodeRef` as part of the state to track DOM nodes for animation.
7. Install `typescript` because it's a needed peer depenency of `prettier-eslint-cli`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/529)


ahangarha (commented):
Thanks for working on upgrading dependencies especially for transition one.

Please fix linter issues.

Also, please consider splitting changes into smaller and relevant commits. Pushing many changes in only one commit makes debugging and reviewing difficult. Many small and relevant commits are better. It helps to know the rationale behind a particular change.

justin808:
@KhaledEmaraDev what's left here?

client/app/bundles/comments/components/CommentBox/CommentList/CommentList.jsx
```diff
@@ -20,57 +20,68 @@ export default class CommentList extends BaseComponent {
...
-        author={$$comment.get('author')}
-        text={$$comment.get('text')}
-      />
-    ));
+    const commentNodes = $$comments.map(($$comment, index) => {
+      const nodeRef = React.createRef(null);
+      return (
+        // `key` is a React-specific concept and is not mandatory for the
+        // purpose of this tutorial. if you're curious, see more here:
+        // http://facebook.github.io/react/docs/multiple-components.html#dynamic-children
```
ahangarha, line 57: Any React developer should know to include unique `key` to lists to avoid issues when interacting with them. We don't need to mention it here.

ahangarha (approved):
Great work 👍🏾 
Thanks to @KhaledEmaraDev for starting and @Yassa-hue for wrapping up.

To me, this PR is done.

I made two comments (which are not blocking)

ahangarha (changes requested):
Good work 👍🏾 

1. After rebasing, we have a new page for stimulus. There, we still get the following warning message:
    image

2. We had a new set of browserlist. After rebase, it is gone.
3. There are some comments from my previous review. Maybe consider applying them with the next commits for wrapping this PR up.

Our only error is related to i18n, which comes from the past and can be fixed in a separate PR.

I don't see any other issues.

client/app/bundles/comments/components/CommentBox/CommentBox.jsx
```diff
@@ -2,8 +2,8 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import Immutable from 'immutable';
 import _ from 'lodash';
-import BaseComponent from 'libs/components/BaseComponent';
 import { injectIntl } from 'react-intl';
+import BaseComponent from 'libs/components/BaseComponent';
```
Yassa-hue, line 6: It's just a linting issue.

client/app/bundles/comments/components/CommentBox/CommentList/CommentList.jsx
```diff
@@ -52,9 +52,6 @@ export default class CommentList extends BaseComponent {
     const commentNodes = $$comments.map(($$comment, index) => {
       const nodeRef = React.createRef(null);
       return (
-        // `key` is a React-specific concept and is not mandatory for the
-        // purpose of this tutorial. if you're curious, see more here:
-        // http://facebook.github.io/react/docs/multiple-components.html#dynamic-children
         <CSSTransition
           key={$$comment.get('id') || index}
```
ahangarha, line 56: I am concerned that this is a good key. Though it works in our case I think it is better to make ids like `comment_123`.

ahangarha (approved):
LGTM 👍🏾

PR #534, opened 2023-07-20
Create Footer component (#499)

### LINKED ISSUE #499: Tab "Simple React" doesn't have footer
image

Yassa-hue:
This commit basically should add a footer to the Simple React page, But The footer logic is used in more than one place so a Footer component is created and called wherever it is needed.

Resolves #499

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/534)


ahangarha (changes requested):
Good move

I see duplicate content at the bottom of the page.

image

What do you think about the Classic Rails page? What to do there?

If we can have the same navigation bar on all pages, we should be able to have a footer too.

Yassa-hue:
Yes, Sorry for forgetting it. I removed the duplicate content.
If we want the footer on every page, I suggest putting it on the application layout with the navigation bar.

client/app/bundles/comments/components/Footer/Footer.module.scss
```diff
@@ -0,0 +1,13 @@
...
+  height: 40px;
+  margin-right: 10px;
+  width: 146px;
+}
+
+.twitterImage {
+  background: url('./images/twitter_64.png') no-repeat;
+  height: 64px;
+  width: 64px;
+}
```
ahangarha, line 13: One blank line at the end of every file

client/app/bundles/comments/components/Footer/Footer.jsx
```diff
@@ -0,0 +1,22 @@
...
+          </h3>
+        </a>
+        <a href="https://twitter.com/railsonmaui">
+          <div className={css.twitterImage} />
+          Rails On Maui on Twitter
+        </a>
+      </div>
+    );
+  }
+}
```
ahangarha, line 22: One blank line at the end of every file

app/views/layouts/application.html.erb
```diff
@@ -24,5 +24,7 @@
      client side hydration -->
 <%= redux_store_hydration_data %>
 
+<%= react_component "Footer" %>
+
```
ahangarha, line 28: Though it doesn't matter in this case, by better to keep `<%= redux_store_hydration_data %>` at the end.

ahangarha (changes requested):
Looks good to me.

There are minor issues to fix. After that, I think this PR is ready to get merged.

app/views/layouts/application.html.erb
```diff
@@ -20,6 +20,10 @@
   <%= yield %>
 </div>
 
+
+<%= react_component "Footer" %>
+
+
```
ahangarha, line 26: I am amazed that our linter doesn't care about multiple blank lines. So I don't make it a blocker. To me, this PR is ready to be merged.

Good job @Yassa-hue 👍🏾

ahangarha (approved):
Good job 👍🏾

ahangarha:
@justin808 Now we have a footer on all pages. Do you have any comments here?

I think we need to update the branding in the footer. Also, the UI/UX can get improved.
I propose to have a separate task for these two.

justin808:
Thanks @Yassa-hue

PR #535, opened 2023-07-25
Replace marked sanitize with sanitize-html (#500)

### LINKED ISSUE #500: Component Comment.jsx is using an insecure method to sanitize input --could suffer XSS atacks
It is using the option `sanitize` of the library `marked`, as seen at: https://github.com/shakacode/react-webpack-rails-tutorial/blob/7feaf746295686f55efde8802039280bdfb25130/client/app/bundles/comments/components/CommentBox/CommentList/Comment/Comment.jsx#L16

However that library warns at <https://marked.js.org/using_advanced#options>:

> Warning: This feature is deprecated and it should NOT be used as it cannot be considered secure.
Instead use a sanitize library, like [DOMPurify](https://github.com/cure53/DOMPurify) (recommended), [sanitize-html](https://github.com/apostrophecms/sanitize-html) or [insane](https://github.com/bevacqua/insane) on the output HTML!

Yassa-hue:
Because the sanitize option in marked is deprecated and not safe This commit replace it with sanitize-html which works well with Server Side Rendering unlike DOMPurify.

Resolves #500

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/535)

### DIFF
--- a/client/app/bundles/comments/components/CommentBox/CommentList/Comment/Comment.jsx
+++ b/client/app/bundles/comments/components/CommentBox/CommentList/Comment/Comment.jsx
@@ -3,6 +3,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 
 import { marked } from 'marked';
+import sanitizeHtml from 'sanitize-html';
 import css from './Comment.module.scss';
 
 export default class Comment extends BaseComponent {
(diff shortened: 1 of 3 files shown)


ahangarha (approved):
Nice small improvement 👍🏾

justin808:
Nice job @Yassa-hue!

PR #539, opened 2023-07-28
Separate tests workflows (#530)

### LINKED ISSUE #530: Separate workflows
Currently, we have all the tests in one unified workflow. Let's separate them into

- Rubocop
- Eslint
- Ruby / e2e

Yassa-hue:
This commit create a separate workflow for each test:
	* rspec tests
	* lint
	* js_tests

Closes #530 

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/539)


lib/tasks/ci.rake
```diff
@@ -13,6 +13,30 @@ if Rails.env.development? || Rails.env.test?
...
+    puts "CI js_tests"
+    puts Rainbow("PASSED").green
+    puts ""
+  rescue StandardError => e
+    puts e.to_s
+    puts Rainbow("FAILED").red
+    puts ""
+    raise(e)
+  end
+
```
ahangarha, line 39: I think these tasks could go under `ci` namespace so that we get these tasks:
- `ci`: runs `ci:all`
- `ci:all` runs all other ci tasks
- `ci:js` runs `js_tests`
- `ci:rspec` runs rspec

ahangarha (changes requested):
Good work 👍🏾 

Please update the PR in a way that we preserve the history of the `test.yml` in the new `rspec_test.yml`.

Also, please check the comment I made about new rake tasks.

ahangarha (approved):
To me, this PR is good.

There are issues with our tests and our CI workflows. I think we should address them in a separate PR.

The objective of this PR was to get the result faster by running parallel workflows and being able to rerun a single failure without needing to run all. This objective is met.

@Judahmeek what do you think?

PR #554, opened 2023-09-15
Add ror binstubs

ahangarha:
Add standard react_on_rails binstubs (`bin/dev` and `bin/dev-static`) for running the server.

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/554)


Judahmeek (approved):
Just one question: where did these bin scripts come from?

Because react_on_rails doesn't have these scripts.

PR #563, opened 2023-10-29
WIP - Improve docker images for faster build and deploy

ahangarha:
Creating a base Dockerfile for this project so that it can be used in `.controlplane/Dockerfile` to speed up the build time.

- This file is based on Rails 7.1 Dockerfile.
- The Docker base image should be built from `Dockerfile_base` and pushed to the Docker hub.
- For deploying on Control Plane, we should use `.controlplane/Dockerfile_from_base`. For this, we need to uncomment the line for the custom Dockerfile path in `.controlplane/controlplane.yml` file.

These new Dockerfiles are made originally with the idea of having a way to make a faster build/deploy process on cpln. We still need the existing Dockerfiles to have more control and flexibility when experimenting with deployment. 

## Result:
Screenshot from 2023-12-09 19-36-43-e

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/563)

### DIFF
--- a/.controlplane/Dockerfile_from_base
+++ b/.controlplane/Dockerfile_from_base
@@ -1,3 +1,7 @@
+# This Dockerfile should be used for a faster image build on top of the base
+# image from Docker hub (shakacode/rwrt-base:latest). To this end, uncomment
+# the line for the custom dockerfile in the controlplane.yml file.
+
 FROM ahangarha/rwrt-base:latest
 
 ENV RAILS_ENV=production
--- a/.controlplane/Dockerfile_from_base
+++ b/.controlplane/Dockerfile_from_base
@@ -2,7 +2,7 @@
 # image from Docker hub (shakacode/rwrt-base:latest). To this end, uncomment
 # the line for the custom dockerfile in the controlplane.yml file.
 
-FROM ahangarha/rwrt-base:latest
+FROM shakacode/rwrt-base:latest
 
 ENV RAILS_ENV=production
 ENV NODE_ENV=production
--- a/.controlplane/Dockerfile_from_base
+++ b/.controlplane/Dockerfile_from_base
@@ -1,0 +1,16 @@
+FROM ahangarha/rwrt-base:latest
+
+ENV RAILS_ENV=production
+ENV NODE_ENV=production
+
+# compiling assets requires any value for ENV of SECRET_KEY_BASE
+ENV SECRET_KEY_BASE=NOT_USED_NON_BLANK
+
+RUN bin/rails react_on_rails:locale
+RUN bin/rails assets:precompile
+
+# add entrypoint
+COPY .controlplane/entrypoint.sh ./
+ENTRYPOINT ["/app/entrypoint.sh"]
+
+CMD ["./bin/rails", "s"]
--- a/.controlplane/Dockerfile_from_base
+++ b/.controlplane/Dockerfile_from_base
@@ -6,6 +6,7 @@ ENV NODE_ENV=production
 # compiling assets requires any value for ENV of SECRET_KEY_BASE
 ENV SECRET_KEY_BASE=NOT_USED_NON_BLANK
 
+RUN yarn res:build
 RUN bin/rails react_on_rails:locale
 RUN bin/rails assets:precompile
 
--- a/.controlplane/Dockerfile_from_base
+++ b/.controlplane/Dockerfile_from_base
@@ -13,4 +13,4 @@ RUN bin/rails assets:precompile
 COPY .controlplane/entrypoint.sh ./
 ENTRYPOINT ["/app/entrypoint.sh"]
 
-CMD ["./bin/rails", "s"]
+CMD ["./bin/rails", "server"]
--- a/.controlplane/entrypoint.sh
+++ b/.controlplane/entrypoint.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash -e
 # Runs before the main command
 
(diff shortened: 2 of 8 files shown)


.controlplane/entrypoint.sh
```diff
@@ -18,8 +18,11 @@ echo " -- Waiting for services"
 wait_for_service $(echo $DATABASE_URL | sed -e 's|^.*@||' -e 's|/.*$||')
 wait_for_service $(echo $REDIS_URL | sed -e 's|redis://||' -e 's|/.*$||')
 
-echo " -- Preparing database"
-rails db:prepare
+# If running the rails server then create or migrate existing database
+if [ "${1}" == "./bin/rails" ] && [ "${2}" == "server" ]; then
+  echo " -- Preparing database"
+  ./bin/rails db:prepare
+fi
```
justin808, line 25: @dzirtusss should this be recommended?

See https://chat.openai.com/share/f7e6d78b-ab5c-4e2f-808c-12c717053613

justin808 (commented):
Why are we adding the root docker file?

How does it relate to the .controlplane/Dockerfile?

@rafaelgomesxyz @ahangarha @Judahmeek

.controlplane/entrypoint.sh
justin808: @ahangarha why is this duplicated in the bin script in this PR?

Dockerfile
```diff
@@ -1,27 +1,72 @@
...
+
+# Install packages needed for deployment
+RUN apt-get update -qq && \
+    apt-get install --no-install-recommends -y curl libvips postgresql-client && \
+    rm -rf /var/lib/apt/lists /var/cache/apt/archives
 
-COPY . ./
+# Copy built artifacts: gems, application
+COPY --from=build /usr/local/bundle /usr/local/bundle
+COPY --from=build /app /app
```
borela, line 60: Is it worth it having a separate build step image vs a single one where you delete the unneeded stuff afterward? You already did some cleaning on line [37](https://github.com/shakacode/react-webpack-rails-tutorial/pull/563/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R37).

Dockerfile
```diff
@@ -1,27 +1,72 @@
...
+# Precompiling assets for production without requiring secret RAILS_MASTER_KEY
+RUN SECRET_KEY_BASE_DUMMY=1 ./bin/rails react_on_rails:locale
+RUN SECRET_KEY_BASE_DUMMY=1 ./bin/rails assets:precompile
+
+# Final stage for app image
+FROM base
+
+# Install packages needed for deployment
+RUN apt-get update -qq && \
+    apt-get install --no-install-recommends -y curl libvips postgresql-client && \
```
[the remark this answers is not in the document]
ahangarha: It comes from Rails new Dockerfile.

Dockerfile
ahangarha: I just followed the Dockerfile proposed by Rails.
It might be still faulty, but it follows the same structure.

.controlplane/entrypoint.sh
ahangarha: The `bin/docker-entrypoint` file comes from new Rails setup.

justin808:
@ahangarha and @Judahmeek how much testing did you do of this?

ahangarha:
@justin808 

> how much testing did you do of this?

I have tested multiple times. The last time, I made sure no image was left on my system and built from scratch (both the base image on the docker hub, and the image to be deployed on cpln)

Have you noticed any issues?

.controlplane/Dockerfile_from_base
```diff
@@ -0,0 +1,17 @@
+FROM ahangarha/rwrt-base:latest
```
justin808, line 1: WTFs:
1. personal name.
2. need comments on why file exists and how to use

ahangarha: The reason behind this was that before pushing this base image to the official docker hub of the company, I could demonstrate deployment. I needed a review on the implementation of this Docker base image. Now I got it and now I can move forward.

.controlplane/entrypoint.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash -e
```
justin808, line 1: why the `-e` flag?

.controlplane/Dockerfile_from_base
```diff
@@ -0,0 +1,21 @@
+# This Dockerfile should be used for a faster image build on top of the base
+# image from Docker hub (shakacode/rwrt-base:latest). To this end, uncomment
+# the line for the custom dockerfile in the controlplane.yml file.
+
+FROM shakacode/rwrt-base:latest
```
justin808, line 5: This is going to confuse people.

ahangarha: I understand.
Perhaps we can move these files into `.shakacode` folder which will be the place for keeping files only related to our development tasks.

.controlplane/entrypoint.sh
ahangarha: I applied it based on the new Rails 7.1 Dockerfile. It is needed for the next changes to run; the `==` operation in the condition.

It is to make the whole Dockerfiles have similar content, making it easier to debug issues in future.

justin808:
Reopen new PR if any of this is still useful.

PR #566, opened 2023-11-19
Show first seven char of commit hash

ahangarha:
This is to follow the same convention of using commit hash so that in one glance we can detect if we are the desired commit or not.

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/566)


justin808:
* No changelog
* @dzirtusss @rafaelgomesxyz can this have backwards compat issues?

PR #575, opened 2023-12-27
Fix the instruction was in a wrong order.

samnang:
There was a step in the wrong order to try to execute cli `cpln` even before it is installed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/575)


ahangarha (approved):
Thanks for the fix.

PR #579, opened 2023-12-30
Staging apps via github actions

justin808:
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/579)


.controlplane/Dockerfile
```diff
@@ -53,6 +54,8 @@ RUN apt-get update -qq && \
 COPY --from=build /usr/local/bundle /usr/local/bundle
 COPY --from=build /app /app
 
+RUN chmod +x /app/.controlplane/*.sh
```
justin808, line 57: @borela is this typical?

borela: This means the files were committed to the repo without +x permission, you can fix that, commit the files again, and remove that chmod.

justin808: I did and it didn't fix the issue.

PR #581, opened 2023-12-31
Improve tests result consistency

ahangarha:
The changes in `rescript_spec.rb` and `turbo_spec.rb` improve our test accuracy.

~The changes in `spec/system/*` fix the test failure when these tests run before others (leading to their failure).~

~I am not sure why we have tests in the `system` directory that should have the type of `feature`, but it fixes our issues. I need to study deeper to understand what is happening under the hood.~

~Should I move these `feature` tests into the `feature` directory, or we can keep them this way for now?~

This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/581)

### DIFF
--- a/spec/stimulus/turbo_spec.rb
+++ b/spec/stimulus/turbo_spec.rb
@@ -39,23 +39,15 @@ describe "with Turbo and Stimulus" do
       visit "/stimulus"
     end
 
-    it "adds a new comment to the page" do
-      fill_in author_field, with: comment.author
-      fill_in text_field, with: comment.text
-      click_button("Post")
-      expect(page).to have_selector "h2", text: comment.author
-    end
-
-    it "comment count increases with successful form submission" do
+    it "adds a new comment to the page and database" do
       initital_comment_count = Comment.all.count
       new_comment_count = initital_comment_count + 1
       fill_in author_field, with: comment.author
       fill_in text_field, with: comment.text
       click_button("Post")
 
-      expect(page).to have_content(comment.author)
-      expect(page).to have_content(comment.text)
-
+      expect(page).to have_css("h2", text: comment.author)
+      expect(page).to have_css("p", text: comment.text)
       expect(Comment.all.count).to equal(new_comment_count)
     end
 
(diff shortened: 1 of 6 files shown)


spec/stimulus/turbo_spec.rb
```diff
@@ -52,27 +52,36 @@
       fill_in author_field, with: comment.author
       fill_in text_field, with: comment.text
       click_button("Post")
-      sleep(1)
+
+      # This will ensure we wait enough for the changes to be applied
+      expect(page).to have_text("")
```
justin808, line 57: @ahangarha, this is a very strange assertion! 

Please add self-review comments for WTF's like this!

CC: @Judahmeek

ahangarha: As I mentioned in the comment above that line, we need to put some delay before getting assertion for comment count change. Before I fixed it with sleep but that is not efficient. This assertion makes it more efficient.

spec/stimulus/turbo_spec.rb
```diff
@@ -52,27 +52,36 @@
       fill_in author_field, with: comment.author
       fill_in text_field, with: comment.text
```
justin808, line 53: make sure the text is some unique value

spec/stimulus/turbo_spec.rb
justin808: expect page to display the new comment in the right format block (not in the input field).

justin808, line 57: image

justin808 (changes requested):

spec/stimulus/turbo_spec.rb
ahangarha: We already had a test to ensure the new comment is displayed on the page. This test was to ensure the comment was added to the database.

Now, we combine them into one test with multiple assertions.

spec/stimulus/turbo_spec.rb
ahangarha: When this test runs, we have a fresh database.

justin808:
@ahangarha see my changes.

PR #609, opened 2024-12-17
Fix i18n error in console

### LINKED ISSUE #559: Fix i18n issue (Missing locale)
Currently we get the following error in the browser console for both server and client side rendering:

```
Error: [@formatjs/intl Error MISSING_DATA] Missing locale data for locale: "en" in Intl.NumberFormat. Using default locale: "en" as fallback. See https://formatjs.io/docs/react-intl#runtime-requirements for more details
```

ihab4real:
Fixes #559

- - -
This change is [Reviewable](https://reviewable.io/reviews/shakacode/react-webpack-rails-tutorial/609)

## Summary by CodeRabbit

- **New Features**
	- Enhanced internationalization support by adding locale data for English, German, Japanese, and Chinese languages.

### DIFF
--- a/client/app/bundles/comments/startup/App.jsx
+++ b/client/app/bundles/comments/startup/App.jsx
@@ -3,6 +3,10 @@ import React from 'react';
 import ReactOnRails from 'react-on-rails';
 
 import NonRouterCommentsContainer from '../containers/NonRouterCommentsContainer.jsx';
+import 'intl/locale-data/jsonp/en';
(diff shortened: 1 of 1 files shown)


Judahmeek:
Optimally, we would check if adding this locale logic increased the bundle size & if it increased the bundle size by too much, then we would look into lazy loading the specific locale chunk we needed.

This is good enough for now though.

